PR #2, opened 2019-04-02
Add common types to kubeflow/common

richardsliu:
Part 1 of https://github.com/kubeflow/tf-operator/issues/960

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/2)


operator/v1/types.go
```diff
@@ -0,0 +1,186 @@
...
+// job, for example how to clean up resources and how long the job can stay
+// active.
+type RunPolicy struct {
+	// CleanPodPolicy defines the policy to kill pods after TFJob is
+	// succeeded.
+	// Default to Running.
+	CleanPodPolicy *CleanPodPolicy `json:"cleanPodPolicy,omitempty"`
+
+	// TTLSecondsAfterFinished is the TTL to clean up jobs (temporary
+	// before kubernetes adds the cleanup controller).
```
jian-he: One nit, we can probably remove this comments now: `temporary before kubernetes adds the cleanup controller` now. 
since the kubernetes has already added [this](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/)

operator/v1/types.go
```diff
@@ -0,0 +1,186 @@
...
+	// determine the behavior when an error occurs:
+	// - 1-127: permanent error, do not restart.
+	// - 128-255: retryable error, will restart the pod.
+	RestartPolicyExitCode RestartPolicy = "ExitCode"
+)
+
+// RunPolicy encapsulates various runtime policies of the distributed training
+// job, for example how to clean up resources and how long the job can stay
+// active.
+type RunPolicy struct {
```
jian-he: I'm up for the idea of a new RunPolicy.  One question:  Once TFJob spec start to use this RunPolicy, it'll break the API compatibility. Is this the expectation ?

operator/v1/types.go
richardsliu: Fixed.

operator/v1/types.go
richardsliu: The plan is to introduce another API version on TFJob. The current version (v1beta2) will not use this repo. 

Also we added support for multiple CRD versions in v0.5, We can also investigate how to use conversion webhooks for API compatibility: https://github.com/kubernetes/enhancements/issues/598. But looks like this feature requires k8s 1.15.

PR #10, opened 2019-04-09
Add more common libraries

### LINKED ISSUE #7: Consider how to refactor pkg/controller.v1beta2/tensorflow
The controller logic under https://github.com/kubeflow/tf-operator/tree/master/pkg/controller.v1beta2/tensorflow and https://github.com/kubeflow/pytorch-operator/tree/master/pkg/controller.v1beta2/pytorch have some code in common. We should consider refactoring the logic there, and move the shared implementation into this repository.

jian-he:
Fixes #7
Extract more common libraries.
In particular, user  needs to provide below func hooks to write their own controller.

The reconcilePods func can be consolidated further, as still some internal logic in the func can be in common. This can be done in next round, IMO
```
	// deleteJobHandler deletes the job
	deleteJobHandler func(job interface{}) error

	// updateJobStatusHandler updates the job status
	updateJobStatusHandler func(job interface{}) error

	// createServiceHandler creates the service
	createServiceHandler func(job interface{}, rtype commonv1.ReplicaType, spec *commonv1.ReplicaSpec, index string) error

	// reconcilePods reconciles the pods for the job
	reconcilePods func(
		job metav1.Object,
		pods []*v1.Pod,
		rtype commonv1.ReplicaType,
		spec *commonv1.ReplicaSpec, rstatus map[string]v1.PodPhase) error
```

@richardsliu @terrytangyuan @johnugeorge 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/10)


richardsliu:
@jian-he Can you add unit tests for your changes? Thanks!

jian-he:
> @jian-he Can you add unit tests for your changes? Thanks!

yep,  can you see if the patch is going a reasonable direction in the meanwhile

job_controller/job_controller.go
```diff
@@ -134,6 +147,22 @@ type JobController struct {
 	// recorder is an event recorder for recording Event resources to the
 	// Kubernetes API.
 	Recorder record.EventRecorder
+
+	// deleteJobHandler deletes the job
+	deleteJobHandler func(job interface{}) error
```
richardsliu: Seems like this is just one line. Can we use a generic way to discover the API to call? Then we can get rid of this.

richardsliu:
@johnugeorge Please also take a look.

job_controller/status.go
```diff
@@ -0,0 +1,116 @@
+package job_controller
```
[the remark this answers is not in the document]
jian-he: You mean the logic in reconcilePods that update the TFjobStatus ? yes, I agree.
we can do it next, as this patch is big enough..

job_controller/job_controller.go
jian-he: I'm not sure how this can be done. 
We plan to write a operator with kube-builder,  with kube-builder, the api for deletion is also different

job_controller/service.go
```diff
@@ -157,3 +159,40 @@ func (jc *JobController) GetServiceSlices(services []*v1.Service, replicas int,
 	}
 	return serviceSlices
 }
+
+// reconcileServices checks and updates services for each given ReplicaSpec.
+// It will requeue the job in case of an error while creating/deleting services.
+func (tc *JobController) reconcileServices(
```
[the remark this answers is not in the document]
jian-he: It has specific implementation such as GetPortFromTFJob() which requires more work, how about do it in next PR

johnugeorge:
JobController Interface should  ultimately contain common utilities and defined common features(PDB, TTL,ActiveDeadline etc)  across all operators.  Just moving functions of tf-operator(in the current form) to JobController might not help our cause though it will reduce duplication.    The scope and logic of current functions(Say in tf-operator) should be first defined well and then move the common logic to JobController.  

eg: ReconcileJobs, ReconcilePods, ReconcileServices  terms look very general and the function scope is not defined. Even if we move these functions to JobController interface, new operator developer cannot use it easily. Instead, we have to break down into scoped functions into common utilities and features and then move common functions to JobController Interface. 

For eg: Currently reconcilePods of tf-operator and pytorch-operator(https://github.com/kubeflow/pytorch-operator/blob/master/pkg/controller.v1beta2/pytorch/pod.go#L60) are different.

jian-he:
The common utilities are not blindly copy-and-paste. It is modified accordingly to accommodate both tf-operator and pytorch-operator.  It indeed contains the common utilities for TTL, activeDeadlines...

In the latest patch, Basically below interfaces are required for custom controller to implement
```
	// DeleteJob deletes the job
	DeleteJob(job interface{}) error

	// UpdateJobStatus updates the job status
	UpdateJobStatus(job interface{}) error

	// CreateNewService creates the service
	CreateNewService(job metav1.Object, rtype commonv1.ReplicaType, spec *commonv1.ReplicaSpec, index string) error

	// ReconcilePods reconciles the pods for the job
	ReconcilePods(
		job metav1.Object,
		pods []*v1.Pod,
		rtype commonv1.ReplicaType,
		spec *commonv1.ReplicaSpec,
		rstatus map[string]v1.PodPhase) error
```

Btw, currently,  we are writing a custom operator using kube-builder to try to use the common libraries. we are experimenting  it in practice to see how to well define the interfaces. I think it can be polished further.

johnugeorge:
Instead of making changes in JobController first and then in operators,  I would suggest making changes in operator first and then make JobController changes accordingly . It will be then easy to separate the common logic

richardsliu:
@jian-he Can we define some common interfaces for the job controller first? Based on what makes sense for a common API, and not based on the current operators' code structure.

For example:
- Create()
- Delete()
- UpdateStatus()
- Reconcile()    
- ???

And make sure all of the functionalities in https://github.com/kubeflow/common/blob/master/operator/v1/types.go can be covered. Then it will help us see how we can refactor the existing operators code.

jian-he:
> Instead of making changes in JobController first and then in operators, I would suggest making changes in operator first and then make JobController changes accordingly . It will be then easy to separate the common logic

Doing it in operator first will need to make changes in tf-operator twice, and also the UTs will need to be changed twice which is more work.
That may also impact the release line of tf-operator ? if the APIs cannot be well-defined in time

jian-he:
> @jian-he Can we define some common interfaces for the job controller first? Based on what makes sense for a common API, and not based on the current operators' code structure.
> 
> For example:
> 
> * Create()
> * Delete()
> * UpdateStatus()
> * Reconcile()
> * ???
> 
> And make sure all of the functionalities in https://github.com/kubeflow/common/blob/master/operator/v1/types.go can be covered. Then it will help us see how we can refactor the existing operators code.

Yep, sounds good.

jian-he:
@richardsliu @johnugeorge 
I did some analysis and think below interfaces can be used for custom controllers.

Custom controllers need to implement below interfaces.
```
	// DeleteJob deletes the job
	DeleteJob(job interface{}) error

	// UpdateJobStatus updates the job status and job conditions
	UpdateJobStatus(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec, jobStatus commonv1.JobStatus) error

	// CreateService creates the service
	CreateService(job interface{}, service *v1.Service) error

	// SetClusterSpec sets the cluster spec for the pod
	SetClusterSpec(job interface{}, podTemplate *v1.PodTemplate, rtype, index string) error

	// CreatePod creates the pod
	CreatePod(job interface{}, podTemplate *v1.PodTemplate) error

```
and inside the controller's reconcile loop, call below method as the entrypoint 
```
func (jc *JobController) reconcileJobs(
       job interface{},
       replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
       jobStatus commonv1.JobStatus,
       runPolicy *commonv1.RunPolicy) error {
```

SimonCqk:
> @richardsliu @johnugeorge
> I did some analysis and think below interfaces can be used for custom controllers.
> 
> Custom controllers need to implement below interfaces.
> 
> ```
> 	// DeleteJob deletes the job
> 	DeleteJob(job interface{}) error
> 
> 	// UpdateJobStatus updates the job status and job conditions
> 	UpdateJobStatus(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec, jobStatus commonv1.JobStatus) error
> 
> 	// CreateService creates the service
> 	CreateService(job interface{}, service *v1.Service) error
> 
> 	// DeleteService deletes the service
> 	DeleteService(job interface{}, service *v1.Service) error
> 
> 	// CreatePod creates the pod
> 	CreatePod(job interface{}, podTemplate *v1.PodTemplate) error
> 
> 	// DeletePod deletes the pod
> 	DeletePod(job interface{}, pod *v1.Pod) error
> 
> 	// SetClusterSpec sets the cluster spec for the pod
> 	SetClusterSpec(job interface{}, podTemplate *v1.PodTemplate, rtype, index string) error
> ```
> and inside the controller's reconcile loop, call below method as the entrypoint
> 
> ```
> func (jc *JobController) reconcileJobs(
>        job interface{},
>        replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
>        jobStatus commonv1.JobStatus,
>        runPolicy *commonv1.RunPolicy) error {
> ```

Is it possible to use `metav1.Object` instead of `interface{}` in `reconcileJobs` and other interface-funcs? Since I've noticed that  `reconcilePods` just use `metav1.Object` as job type, and when step into `reconcileJobs` there will be a type assertion from `interface{}` to `metav1.Object`. Is there some extra considerations ?

jian-he:
> > @richardsliu @johnugeorge
> > I did some analysis and think below interfaces can be used for custom controllers.
> > Custom controllers need to implement below interfaces.
> > ```
> > 	// DeleteJob deletes the job
> > 	DeleteJob(job interface{}) error
> > 
> > 	// UpdateJobStatus updates the job status and job conditions
> > 	UpdateJobStatus(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec, jobStatus commonv1.JobStatus) error
> > 
> > 	// CreateService creates the service
> > 	CreateService(job interface{}, service *v1.Service) error
> > 
> > 	// DeleteService deletes the service
> > 	DeleteService(job interface{}, service *v1.Service) error
> > 
> > 	// CreatePod creates the pod
> > 	CreatePod(job interface{}, podTemplate *v1.PodTemplate) error
> > 
> > 	// DeletePod deletes the pod
> > 	DeletePod(job interface{}, pod *v1.Pod) error
> > 
> > 	// SetClusterSpec sets the cluster spec for the pod
> > 	SetClusterSpec(job interface{}, podTemplate *v1.PodTemplate, rtype, index string) error
> > ```
> > and inside the controller's reconcile loop, call below method as the entrypoint
> > ```
> > func (jc *JobController) reconcileJobs(
> >        job interface{},
> >        replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
> >        jobStatus commonv1.JobStatus,
> >        runPolicy *commonv1.RunPolicy) error {
> > ```
> 
> Is it possible to use `metav1.Object` instead of `interface{}` in `reconcileJobs` and other interface-funcs? Since I've noticed that `reconcilePods` just use `metav1.Object` as job type, and when step into `reconcileJobs` there will be a type assertion from `interface{}` to `metav1.Object`. Is there some extra considerations ?

There are places that require "runtime.Object" type, e.g. the "Recorder.Event" method: 
[link](https://github.com/kubeflow/common/pull/10/files#diff-143f060d3d26809c19a8a0450c1b6692R439)

jian-he:
Since it's too big to do everything in one patch,  opened #12 for creating the interfaces first.
We can implement the individual pieces gradually.

job_controller/job_controller.go
```diff
@@ -3,7 +3,13 @@ package job_controller
 import (
 	"errors"
 	"fmt"
+	commonv1 "github.com/kubeflow/common/operator/v1"
```
gaocegege: Could we place third-party deps after the first group? Ref https://github.com/golang/go/wiki/CodeReviewComments#imports

job_controller/service.go
```diff
@@ -15,13 +15,15 @@ package job_controller
...
-
+	commonv1 "github.com/kubeflow/common/operator/v1"
+	"github.com/kubeflow/common/util"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/kubernetes/pkg/controller"
+	"strconv"
+	"strings"
```
gaocegege: Same here

util/k8sutil/k8sutil.go
```diff
@@ -15,6 +15,7 @@
 package k8sutil
 
 import (
+	commonv1 "github.com/kubeflow/common/operator/v1"
```
gaocegege: Same here

jian-he:
I experimented more and found the current set of interfaces are still not enough... below is the newest set of interfaces I got. 
```
	// DeleteJob deletes the job
	DeleteJob(job interface{}) error

	// UpdateJobStatus updates the job status objects with new job conditions based on replica statuses
	// e.g. Check whether worker 0 is exited without error and set job status
	UpdateJobStatus(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
		jobStatus *commonv1.JobStatus, restart bool) error

	// Update Job Status in Api server
	UpdateJobStatusInApiServer(job interface{}) error

	// CreateService creates the service, note that the underlying implementation is supposed
	// to add the job to the service's ownerReference
	CreateService(job interface{}, service *v1.Service) error

	// DeleteService deletes the service
	DeleteService(job interface{}, service *v1.Service) error

	// CreatePod creates the pod
	CreatePod(job interface{}, podTemplate *v1.PodTemplateSpec) error

	// DeletePod deletes the pod
	DeletePod(job interface{}, pod *v1.Pod) error

	// SetClusterSpec sets the cluster spec for the pod
	SetClusterSpec(job interface{}, podTemplate *v1.PodTemplateSpec, rtype, index string) error

	// Returns if this replica is a master role
	IsMasterRole(replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec, rtype commonv1.ReplicaType, index int) bool
``` 
I uploaded a newest patch to demo the idea -  it is better to understand the interfaces with real code context. 
Please check if the interfaces makes sense, thanks.
In particular, I'm still looking for a way to combine the UpdateJobStatusInApiServer and UpdateJobStatus

@richardsliu @gaocegege @terrytangyuan

util/k8sutil/k8sutil.go
jian-he: yep, will fix

job_controller/job_controller.go
```diff
@@ -320,3 +369,201 @@ func (jc *JobController) resolveControllerRef(namespace string, controllerRef *m
 	}
 	return job
 }
+
+// reconcileJobs checks and updates replicas for each given ReplicaSpec.
+// It will requeue the job in case of an error while creating/deleting pods/services.
+func (jc *JobController) reconcileJobs(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
```
johnugeorge: ReconcileJobs is misleading in JobController

jian-he: Do you mean the name is misleading ? 
This is the entrypoint that is supposed to be called by TFJob etc. reconcile loop.

johnugeorge: I meant, Reconcile function looks very general on a whole. When we move into the common code, its interface should be well defined. Else, new developers might not be able to use it.

job_controller/job_controller.go
```diff
@@ -60,6 +73,42 @@ type ControllerInterface interface {
...
 	GetJobFromAPIClient(namespace, name string) (metav1.Object, error)
+
+	// Returns the default container in pod. e.g. tensorflow
+	GetDefaultContainerName() string
+
+	// Returns the default container port name. e.g. tfjob-port
+	GetDefaultContainerPortName() string
+
+	// DeleteJob deletes the job
+	DeleteJob(job interface{}) error
```
johnugeorge: similarl to DeleteJob, is there a CreateJob ?

richardsliu:
@johnugeorge Can you review this PR instead? https://github.com/kubeflow/common/pull/12

As mentioned in https://github.com/kubeflow/common/pull/10#issuecomment-482300460, we are splitting this work into several PRs, starting with the interface.

job_controller/job_controller.go
jian-he: Actually, reconcileJobs currently is not an interface. it provides concrete implementations.
The custom operators has to somehow invoke (or be invoked by) the library, and that is this entrypoint method.
We should write proper documentations and examples for how to use it.
Do you have any suggestions for how to construct this?

job_controller/job_controller.go
jian-he: CreateJob is done by the user, when creating a CRD (i.e. TFJob). operators will not create its own CRDs

job_controller/job_controller.go
```diff
@@ -60,6 +73,42 @@ type ControllerInterface interface {
...
+
+	// UpdateJobStatus updates the job status objects with new job conditions based on replica statuses
+	// e.g. Check whether worker 0 is exited without error and set job status
+	UpdateJobStatus(job interface{}, replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
+		jobStatus *commonv1.JobStatus, restart bool) error
```
SimonCqk: I'm a little confused with `jobStatus` passed inhere, is this a new job status or just the inside attribution of `job` ?

jian-he: It's the inner field of the job.  we could get the fields by typecasting the job but that will assume the XXXJob must embed the common.JobStatus which may or may not be true. XXXJob could have its own XXXJobStatus which embeds common.JobStatus.

jian-he:
another splitted PR https://github.com/kubeflow/common/pull/14 is opened

gaocegege:
Please solve the conflicts, thanks.

jian-he:
This giant patch is here for reference purpose.
It'll be splitted up individually

gaocegege:
OK, SGTM

johnugeorge:
Do we still need this PR ?

jian-he:
No we don't, let me close this.

PR #12, opened 2019-04-11
Define interfaces for common job operators

jian-he:
Creating the interfaces for custom controllers to implement, that is:

```
	// DeleteJob deletes the job
	DeleteJob(job interface{}) error

	// UpdateJobStatus updates the job status and job conditions
	UpdateJobStatus(job interface{}, replicas map[common.ReplicaType]*common.ReplicaSpec, jobStatus common.JobStatus) error

	// CreateService creates the service
	CreateService(job interface{}, service *v1.Service) error

	// DeleteService deletes the service
	DeleteService(job interface{}, service *v1.Service) error

	// CreatePod creates the pod
	CreatePod(job interface{}, podTemplate *v1.PodTemplate) error

	// DeletePod deletes the pod
	DeletePod(job interface{}, pod *v1.Pod) error

	// SetClusterSpec sets the cluster spec for the pod
	SetClusterSpec(job interface{}, podTemplate *v1.PodTemplate, rtype, index string) error
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/12)

@richardsliu @johnugeorge @terrytangyuan

### DIFF
--- a/job_controller/job_controller.go
+++ b/job_controller/job_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	common "github.com/kubeflow/common/operator/v1"
 	"github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
 	kubebatchclient "github.com/kubernetes-sigs/kube-batch/pkg/client/clientset/versioned"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


job_controller/job_controller.go
```diff
@@ -60,6 +61,27 @@ type ControllerInterface interface {
 
 	// Returns the Job from API server
 	GetJobFromAPIClient(namespace, name string) (metav1.Object, error)
+
+	// DeleteJob deletes the job
+	DeleteJob(job interface{}) error
```
gaocegege: maybe runtime.Object is better?

johnugeorge: And CreateJob missing?

jian-he: Some methods require both "metav1.Object" (for things like GetName,GetNamespace) and "runtime.Object" (for ".Recorder.Eventf")types, so thought it's better to be make it consistent to be interface{}

gaocegege: SGTM

gaocegege: Yeah, same question with @johnugeorge 

Can we define CreateJob here?

jian-he: CreateJob is actually done by the user when creating a CRD (i.e. TFJob). I think operators will not create its own CRDs ?

gaocegege: Yeah, sure, I think so, thanks.

Issue #11, opened 2019-04-11
Add SchedulingPolicy for the common operator


k82cn:
Add `.spec.runtime.schedulingPolicy` for common operator to use batch scheduler.

gaocegege:
From https://github.com/kubeflow/common/pull/13#issuecomment-483880849

@k82cn 
> Sorry for post update: there's a "feature" in upstream, named SchedulingPolicy, which is used to define whether a scheduler feature can be enabled, e.g. toleration. Do we need to have another term for that, e.g. SchedulingSpec?
>
> xref kubernetes/community#1937

@gaocegege 
> Thanks for the information, Kubeflow requires k8s 1.8, does this feature exist in 1.8?

@k82cn 
 > > Thanks for the information, Kubeflow requires k8s 1.8, does this feature exist in 1.8?
> 
> nop, but this may make other contributor confuse.
> 
> BTW, I think it should be client-go 1.8, k8s 1.11 :)

@jian-he @johnugeorge @richardsliu @terrytangyuan 

WDYT

PR #13, opened 2019-04-13
Added schedulingPolicy api.

k82cn:
Signed-off-by: Da K. Ma <klaus1982.cn@gmail.com>

part of #11

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/13)

### DIFF
--- a/operator/v1/types.go
+++ b/operator/v1/types.go
@@ -179,7 +179,13 @@ type RunPolicy struct {
 	// +optional
 	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
 
-        // TODO - Implement Scheduling policy.
-	// See https://github.com/kubeflow/tf-operator/issues/916#issuecomment-458729706
-        // SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
+	// SchedulingPolicy defines the policy related to scheduling, e.g. gang-scheduling
+	// +optional
+	SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
+}
+
+// SchedulingPolicy encapsulates various scheduling policies of the distributed training
+// job, for example `minAvailable` for gang-scheduling.
+type SchedulingPolicy struct {
+	MinAvailable *int32 `json:"minAvailable,omitempty"`
 }


k82cn:
Add api firstly, will add implementation later.

k82cn:
> Could we support advanced use cases? For example, we want to run PS in different machines and place workers around PS.

sure, we have implemented this internally; will publish it later.

johnugeorge:
You also discussed about adding queue name.  Is it planned?

k82cn:
yes, queue is a key feature in the pipeline. I will add integration when it is stable, e.g. `QueueName`.

operator/v1/types.go
```diff
@@ -179,7 +179,13 @@ type RunPolicy struct {
...
-        // TODO - Implement Scheduling policy.
-	// See https://github.com/kubeflow/tf-operator/issues/916#issuecomment-458729706
-        // SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
+	// SchedulingPolicy defines the policy related to scheduling, e.g. gang-scheduling
+	// +optional
+	SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
+}
+
+// SchedulingPolicy encapsulates various scheduling policies of the distributed training
+// job, for example `minAvailalbe` for gang-scheduling.
```
richardsliu: Nit: typo here for minAvailable

gaocegege:
@k82cn Could you please fix the typo, then I think we could merge it. Thanks

k82cn:
fixed typo :)

k82cn:
Sorry for post update: there's a "feature" in upstream, named `SchedulingPolicy`, which is used to define whether a scheduler feature can be enabled, e.g. toleration. Do we need to have another term for that, e.g. `SchedulingSpec`?

xref https://github.com/kubernetes/community/pull/1937

gaocegege:
@k82cn Thanks for the information, Kubeflow requires k8s 1.8, does this feature exist in 1.8?

k82cn:
> Thanks for the information, Kubeflow requires k8s 1.8, does this feature exist in 1.8?

nop, but this may make other contributor confuse.

BTW, I think it should be client-go 1.8, k8s 1.11 :)

PR #16, opened 2019-04-17
Added .gitignore file

terrytangyuan:
Added some common file patterns that need to be ignored. Some users might still want to generate `/vendor` folder via `go build -mod=vendor` even though Go Modules doesn't generate it by default so I included it here as well.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/16)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,68 @@
+# IDE related
 .idea/
 .vscode/
-.DS_Store
 .swp
+
+# Some users might still want to generate this via `go build -mod=vendor`
+# even though Go Modules doesn't generate it by default
 vendor/
+
+# Items below are adopted from https://github.com/github/gitignore/blob/master/Go.gitignore
+
+## Binaries for programs and plugins
+*.exe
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,5 @@
+.idea/
+.vscode/
+.DS_Store
+.swp
+vendor/
```
gaocegege: Could we reuse the ignore files in https://github.com/github/gitignore?

terrytangyuan: Yes I think so. Do you mean this one specifically? https://github.com/github/gitignore/blob/master/Go.gitignore
Let me know if I miss anything else that I should be adding.

gaocegege: Yeah, https://github.com/github/gitignore/blob/master/Global/Linux.gitignore and https://github.com/github/gitignore/blob/master/Global/macOS.gitignore also can be added

terrytangyuan: Thanks. Added.

PR #29, opened 2019-04-25
Unify the label keys

jian-he:
Since we have a common operator, and think the common label keys can also be unified,  so that the custom operator will not need to implement these

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/29)

### DIFF
--- a/job_controller/job_controller.go
+++ b/job_controller/job_controller.go
@@ -45,23 +45,8 @@ type ControllerInterface interface {
 	// Returns the GroupVersion of the API
 	GetAPIGroupVersion() schema.GroupVersion
 
-	// Returns the Group Name(key) in the labels of the job
-	GetGroupNameLabelKey() string
-
-	// Returns the Job Name(key) in the labels of the job
-	GetJobNameLabelKey() string
-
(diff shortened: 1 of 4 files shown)


johnugeorge:
if same labels are used for all job operators, Will the pod label key-value collide for jobs with same name in the same namespace?  (When pods are listed)

jian-he:
> if same labels are used for all job operators, Will the pod label key-value collide for jobs with same name in the same namespace? (When pods are listed)

I think the pod can not be created in the first place, because the pod name needs to be unique, see how the pod name is [generated](https://github.com/kubeflow/tf-operator/blob/70cafb1765237af21fd1d7df4d052056850f404b/pkg/common/jobcontroller/util.go#L24)

we can add a label "job-type=tensorflow" if needed.

Today the tf-operator [implementation](https://github.com/kubeflow/tf-operator/blob/70cafb1765237af21fd1d7df4d052056850f404b/pkg/common/jobcontroller/pod.go#L177)  always list all pods regardless job types, which I think can be improved.

richardsliu:
/lgtm
/approve
+1 for adding the job-type label later.

PR #30, opened 2019-04-25
Make UpdateJobConditions public to be used by custom operators

jian-he:
UpdateJobConditions can be used as the util method by custom operators when updating the JobStatus in the UpdateJobStatus api

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/30)


jian-he:
sounds good, which package/file should we put ? 
util/util.go ?

gaocegege:
SGTM. job_controller/util.go

jian-he:
I ended up moving the util methods to util/status.go.  Please check if making sense

jian-he:
this patch only moves the code from one to another and should not affect test coverage

PR #31, opened 2019-04-26
make createNewPod public for custom operators.

SimonCqk:
For now `createNewPod` is a private method and used in `ReconcilePods`, but for customized operators, the pod reconciliation is slightly different that we may asked to maintain and update other status, so we have to override the original `ReconcilePods` in common lib, then `createNewPod` can not be reused, so I propose to make this method public.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/31)

### DIFF
--- a/job_controller/pod.go
+++ b/job_controller/pod.go
@@ -315,7 +315,7 @@ func (jc *JobController) ReconcilePods(
 
 			// check if this replica is the master role
 			masterRole = jc.Controller.IsMasterRole(replicas, rtype, index)
-			err = jc.createNewPod(job, rt, strconv.Itoa(index), spec, masterRole, replicas)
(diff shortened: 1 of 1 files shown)


jian-he:
could you give a concrete example on which part of the current reconcilePod cannot apply

johnugeorge:
Instead of fixing each function by function(See #30 ),  shouldn't the restructuring be done more systematically?

@richardsliu

SimonCqk:
> Instead of fixing each function by function(See #30 ), shouldn't the restructuring be done more systematically?
> 
> @richardsliu

Do agree, there may other cases to be ensured.

richardsliu:
@SimonCqk What is the plan for this PR? Are we creating a util class as mentioned in https://github.com/kubeflow/common/pull/30?

SimonCqk:
@richardsliu It seems this PR can be closed.

PR #34, opened 2019-05-05
set podInformer/serviceInformer in NewJobController

sperlingxx:
Binding `podInformer` and `serviceInformer` onto jobController when we initialize a jobController by calling `NewJobController`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/34)

### DIFF
--- a/job_controller/job_controller.go
+++ b/job_controller/job_controller.go
@@ -191,6 +191,7 @@ func NewJobController(
 		EnableGangScheduling:     enableGangScheduling,
 	}
 
+
 	jc := JobController{
 		Controller:         controllerImpl,
 		Config:             jobControllerConfig,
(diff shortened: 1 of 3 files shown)


jian-he:
After https://github.com/kubeflow/common/pull/36 the usage of podLister/serviceLister are removed, do we still need this change ?
If projects use kube-builder to generate the skeleton code, then I think the raw podInformer/serviceInformer structure  won't be needed too.

gaocegege:
We do not use kubebuilder now. Thus we need the informer to watch resources in apiserver. But I am not sure if listers are needed here.

gaocegege:
Did we decided to use kubebuilder in tf-operator and pytorch-operator? @jian-he 

I am not aware of it.

jian-he:
> Did we decided to use kubebuilder in tf-operator and pytorch-operator? @jian-he
no, we didn't. Just want to mention this scenario.

Yes, we still need to use informer etc. Just think should that reside  in specific job controller's own implementation (the ones that require the informer structure) or still put it in the common API.

gaocegege:
@jian-he Oh gotcha. Yeah it is the problem. Personally, I prefer keeping it in specific job controller. Or we will introduce something useless for kubebuilder use cases.

johnugeorge:
agree with @gaocegege

terrytangyuan:
Agreed. Then I think we can close this. WDYT? @sperlingxx

sperlingxx:
> Agreed. Then I think we can close this. WDYT? @sperlingxx

I close this now.

PR #36, opened 2019-05-09
Remove the podControl and serviceControl interfaces

### LINKED ISSUE #33: PodControl/ServiceControl interfaces are duplicated with the JobController create/delete Pod/Service Interface
https://github.com/kubeflow/common/blob/master/job_controller/job_controller.go#L68-L77
https://github.com/kubeflow/common/blob/master/job_controller/job_controller.go#L111-L114

Problem
Above two interfaces are both used for underlying create/delete Pod/Service implementations, which are duplicated.  

Goal
- we should keep one of them
- only in one place, i.e. ControllerInterface.

Solution
1)   Originally thought we can remove the Pod/ServiceControl interface, but there are some code relying on it like [here](https://github.com/kubeflow/common/blob/master/job_controller/pod.go#L226) used for the parameter.  - we could rewrite these to remove the dependency though

2) Keep the Pod/Service Interface, and move it to one place , i.e. the ControllerInterface. I also realized some implementations are directly copied from Kubernetes [utils](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L520), wondering why not just use that. 

opinions?
@richardsliu @gaocegege @terrytangyuan

jian-he:
Fixes #33 

Removed the PodControl/ServiceControl interfaces.
Changed the associated logic accordingly

@richardsliu @gaocegege @johnugeorge 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/36)

### DIFF
--- a/job_controller/job.go
+++ b/job_controller/job.go
@@ -93,14 +93,14 @@ func (jc *JobController) ReconcileJobs(
 
 	oldStatus := jobStatus.DeepCopy()
 
-	pods, err := jc.GetPodsForJob(metaObject)
+	pods, err := jc.Controller.GetPodsForJob(job)
 
 	if err != nil {
 		log.Warnf("GetPodsForJob error %v", err)
 		return err
 	}
 
-	services, err := jc.GetServicesForJob(metaObject)
+	services, err := jc.Controller.GetServicesForJob(job)
 
 	if err != nil {
 		log.Warnf("GetServicesForJob error %v", err)
--- a/job_controller/job_controller.go
+++ b/job_controller/job_controller.go
@@ -55,6 +55,14 @@ type ControllerInterface interface {
 	// Returns the Job from API server
 	GetJobFromAPIClient(namespace, name string) (metav1.Object, error)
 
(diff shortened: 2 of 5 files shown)


job_controller/job.go
```diff
@@ -93,14 +93,14 @@ func (jc *JobController) ReconcileJobs(
 
 	oldStatus := jobStatus.DeepCopy()
 
-	pods, err := jc.GetPodsForJob(metaObject)
+	pods, err := jc.Controller.GetPodsForJob(job)
```
gaocegege: Do we need to specify `Controller` here? Interface embedding will do it for us, I think.

jian-he: that's due to the [controller field](https://github.com/kubeflow/common/blob/c99eb21e464ea0365b77a1172f5e7e8c4468cb11/job_controller/job_controller.go#L117) is explicitly used in the struct. 

this was there originally, we can remove it in a separate PR.

gaocegege: Gotcha. SGTM

PR #45, opened 2019-08-23
add job priority for kube-batch scheduling #141

YesterdayxD:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/45)

### DIFF
--- a/job_controller/api/v1/types.go
+++ b/job_controller/api/v1/types.go
@@ -188,4 +188,8 @@ type RunPolicy struct {
 // job, for example `minAvailable` for gang-scheduling.
 type SchedulingPolicy struct {
 	MinAvailable *int32 `json:"minAvailable,omitempty"`
+
+	//add PriorityClassName
+	//PriorityClassName is a type of k8s resource.(kubectl get priorityclass)
+	PriorityClassName *string `json:"priorityClassName,omitempty"`
 }
--- a/job_controller/api/v1/types.go
+++ b/job_controller/api/v1/types.go
@@ -189,7 +189,6 @@ type RunPolicy struct {
 type SchedulingPolicy struct {
 	MinAvailable *int32 `json:"minAvailable,omitempty"`
 
-	//add PriorityClassName
 	//PriorityClassName is a type of k8s resource.(kubectl get priorityclass)
 	PriorityClassName *string `json:"priorityClassName,omitempty"`
 }
--- a/job_controller/job.go
+++ b/job_controller/job.go
@@ -1,6 +1,7 @@
 package job_controller
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -40,7 +41,6 @@ func (jc *JobController) deletePodsAndServices(runPolicy *apiv1.RunPolicy, job i
 	}
 	return nil
 }
-// TTL means time to live
 func (jc *JobController) cleanupJobIfTTL(runPolicy *apiv1.RunPolicy, jobStatus apiv1.JobStatus, job interface{}) error {
 	currentTime := time.Now()
 	metaObject, _ := job.(metav1.Object)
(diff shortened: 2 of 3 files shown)


job_controller/job.go
```diff
@@ -191,6 +191,16 @@ func (jc *JobController) ReconcileJobs(
 		return jc.Controller.UpdateJobStatusInApiServer(job, &jobStatus)
 	}
 
+	if jc.Config.EnableGangScheduling {
+		minAvailableReplicas := getTotalReplicas(replicas)
+		priorityClassName := getPriorityClassName(runPolicy)
+		//_, err := pc.SyncPodGroup(job, minAvailableReplicas)
```
merlintang: why keep this ?

YesterdayxD: I reference the code about EnableGangScheduling of other operators and add the priority‘s’attribute and function.

job_controller/api/v1/types.go
```diff
@@ -188,4 +188,8 @@ type RunPolicy struct {
 // job, for example `minAvailable` for gang-scheduling.
 type SchedulingPolicy struct {
 	MinAvailable *int32 `json:"minAvailable,omitempty"`
+
+	//add PriorityClassName
```
gaocegege: We can remove this line

job_controller/job.go
```diff
@@ -40,7 +40,7 @@ func (jc *JobController) deletePodsAndServices(runPolicy *apiv1.RunPolicy, job i
 	}
 	return nil
 }
-
+// TTL means time to live
```
gaocegege: This line, too.

job_controller/job.go
```diff
@@ -299,3 +309,14 @@ func (jc *JobController) cleanupJob(runPolicy *apiv1.RunPolicy, jobStatus apiv1.
 	jc.WorkQueue.AddRateLimited(key)
 	return nil
 }
+func getPriorityClassName(runPolicy *apiv1.RunPolicy) string {
+	priorityClassName := *runPolicy.SchedulingPolicy.PriorityClassName
```
gaocegege: Runtime error here

job_controller/job.go
```diff
@@ -299,3 +309,14 @@ func (jc *JobController) cleanupJob(runPolicy *apiv1.RunPolicy, jobStatus apiv1.
...
+	return priorityClassName
+}
+func getTotalReplicas(replicas map[apiv1.ReplicaType]*apiv1.ReplicaSpec) int32 {
+	jobReplicas := int32(0)
+	for _, r := range replicas {
+		jobReplicas += *r.Replicas
```
gaocegege: I think we should check if r.Replicas is nil here, too.

YesterdayxD: I got it

YesterdayxD:
I try to correct it

terrytangyuan:
Hmm I just asked the author of this PR to submit the changes here: https://github.com/kubeflow/mpi-operator/pull/141

Are you in the same team? It would be better if you can send a note before submitting the PR to avoid redundant work...

terrytangyuan:
cc @4everming

YesterdayxD:
@terrytangyuan Yes, we are in same team. I will be more careful next time. Thank you for reminding me.
 @4everming

gaocegege:
@terrytangyuan Since the MPI operator does not use the common operator now, should we keep the similar logic in these two repos?

terrytangyuan:
Yes I think it’s fine to do that for now. Though the types like RunPolicy is used in MPI operator so that should be reused.

gaocegege:
Yeah, I think so. 

BTW @YesterdayxD @4everming

Really appreciate it if you could open an issue to illustrate why we need the change and how to implement it in several operators.

Personally, I know that there is a queue in PodGroup and we can set it to define the priority. And the default value is `default`. But I think we should have a doc or issue to keep track of the change. Maybe we could cc k82cn@ to request a review, too.

YesterdayxD:
ok，I will open an issue about the PR

k82cn:
that's great, I'd like to contribute too :)

4everming:
> Yeah, I think so.
> 
> BTW @YesterdayxD @4everming
> 
> Really appreciate it if you could open an issue to illustrate why we need the change and how to implement it in several operators.
> 
> Personally, I know that there is a queue in PodGroup and we can set it to define the priority. And the default value is `default`. But I think we should have a doc or issue to keep track of the change. Maybe we could cc k82cn@ to request a review, too.

Sure thing. Thanks a lot.

gaocegege:
https://github.com/kubeflow/common/issues/46

@4everming and @YesterdayxD are working on a proposal, @ me and k82cn@ if you think it is ready to review.

terrytangyuan:
Closing this due to inactivity. Feel free to re-open if this is picked up again.

PR #49, opened 2019-12-07
Update openapi tag for SDK generating

jinchihe:
In the PR https://github.com/kubeflow/tf-operator/pull/1103, we generate SDK using openapi  -> swagger ways, tfjob depends on common repo, the PR is current the openapi generate flag so that the current types can be involved. See more in https://github.com/kubeflow/tf-operator/pull/1103.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/49)


jinchihe:
/assign @gaocegege 
Could you please take a look, thanks.

PR #51, opened 2020-03-08
Move JobController struct to job_controller package

Jeffwan:
It doesn't make sense to have JobController struct definition in API definition. `JobController` can be changed and we also provide  `NewJobController` method for user to use, that means controller could be hided from users and implementation may changed even for same API version . It doesn't have to be changed along with api changes. 

This also makes dependency graph clean for `api/v1` package.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/51)


Jeffwan:
https://github.com/kubeflow/common/pull/55 includes the changes in this PR. I will close this one

Issue #52, opened 2020-03-08
Deprecate glog and use logger in commonutil instead


Jeffwan:
glog has been deprecated in the kubernetes community for a few reasons.
I notice it's still being used in the common project and I would be better to clean up all references and migrate to existing log solutions.

/improvement

Jeffwan:
PR merged and this can be closed

PR #53, opened 2020-03-08
Deprecate glog and move to commonutil logger

Jeffwan:
Address #52 

1. Remove all glog references and use commonutil logger instead.
2. Create method LoggerForService in logger
3. Clean up go.mod and go.sum

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/common/53)


Jeffwan:
> Thanks! Have other operators removed the use of glog yet? If not, would this break existing 3rd party logging service by using different logging frameworks?

@terrytangyuan Tf-operator and mpi-operator still has few references. I will file follow up PRs to migrate them either to klog or logrus later. This will not break existing logging service. The situation is we have multiple logs solutions and this PR helps to use unified one across the project.  (In this project, logrus)

Jeffwan:
Fix an edge case in the test and please have another look.

Issue #56, opened 2020-03-25
Make training container port optional


Jeffwan:
Instead of using fixed port to communicate with each other, some frameworks randomly pick up port by worker and use auto-discovery to find entire topology in consensus store. 

Currently, common force user to implement `GetDefaultContainerPortName` and `GetDefaultContainerPortNumber` and use it in following logic.  

https://github.com/kubeflow/common/blob/64f943084a050ac84b50ce4f49b6b6e085b247e6/pkg/controller.v1/common/service.go#L174

If job doesn't have port, it will return an error. 

In order to make this change, we can make this optional to meet the case I described above.

terrytangyuan:
Sounds good to me.

Jeffwan:
PR has been merged and I will close the issue

PR #57, opened 2020-03-25
Make container port optional

Jeffwan:
Address #56 

- Instead of using fixed port to communicate with each other, some frameworks randomly pick up port by worker and use auto-discovery to find entire topology.
- In that case, we only need pod-to-pod communication and don’t need to specify port in headless service.
- This give the flexibility for framework which doesn’t rely on fixed port.

I've tested in a controller using this common.


Jeffwan:
Rebase the master.

PR #62, opened 2020-04-09
Import volcano api

hzxuzhonghu:
Volcano is now being in the process of becoming a CNCF sandbox project. And many operators like tf-operator has already migrated to volcano scheduler

https://github.com/cncf/toc/pull/318


Jeffwan:
`Vendor` has been deprecated. I will file another PR to delete it. In this case, I don't think you need to vendor dependencies.

go.mod
```diff
@@ -1,41 +1,31 @@
...
 	gopkg.in/square/go-jose.v2 v2.3.0 // indirect
 	k8s.io/api v0.15.10
-	k8s.io/apimachinery v0.15.10
-	k8s.io/client-go v0.15.10
+	k8s.io/apimachinery v0.16.9-beta.0
+	k8s.io/client-go v0.16.9-beta.0
 	k8s.io/code-generator v0.15.10
 	k8s.io/kube-openapi v0.0.0-20190228160746-b3a7cee44a30
-	k8s.io/kubernetes v1.15.10
+	k8s.io/kubernetes v1.18.0
```
Jeffwan, line 21: Can we try to minimize the dependency changes? Another thing is `k8s.io/Kubernetes` version doesn't match with apimachinery and client-go

pkg/controller.v1/common/job_controller.go
```diff
@@ -66,8 +66,8 @@ type JobController struct {
 	// KubeClientSet is a standard kubernetes clientset.
 	KubeClientSet kubeclientset.Interface
 
-	// KubeBatchClientSet is a standard kube-batch clientset.
-	KubeBatchClientSet kubebatchclient.Interface
```
Jeffwan, line 70: I think my question is, do we want to have some backward compatibility or do we want to support different backends  in case some one want to use `kube-batch` instead of `valcano`?

go.mod
k82cn: +1 to keep current version :) Is there any reason for that?

hzxuzhonghu: Actually we import 1.15.10, you can see the replace below. This is autogenrerated

hzxuzhonghu:
>Vendor has been deprecated. I will file another PR to delete it. In this case, I don't think you need to vendor dependencies. This project has been migrate to go mod already.

Agree, i just follow the previous way

pkg/controller.v1/common/job_controller.go
hzxuzhonghu: Stand on my own point, as volcano is more actively maintained, it is better to migrate to it in the next release

go.mod
```diff
@@ -51,6 +41,7 @@ replace (
 	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.15.10
 	k8s.io/kubectl => k8s.io/kubectl v0.15.11-beta.0
 	k8s.io/kubelet => k8s.io/kubelet v0.15.10
+	k8s.io/kubernetes => k8s.io/kubernetes v1.15.10
```
hzxuzhonghu, line 44: @Jeffwan please see here

terrytangyuan:
> `Vendor` has been deprecated. I will file another PR to delete it. In this case, I don't think you need to vendor dependencies. This project has been migrate to go mod already.

Yes let's delete it.

pkg/controller.v1/common/job_controller.go
terrytangyuan: How much efforts would there be to support multiple backends? Would it be possible to make it pluggable?

go.mod
Jeffwan: Em. It must be some libraries requires high version?  I notice volcano  go.mod requires 1.16.9 https://github.com/volcano-sh/volcano/blob/9c314bd9e262413be6a902ae1dc6dbe133732ca5/go.mod#L49-L54

which matches apimachinery and client-go version here. `k8s.io/kubernetes` must be from other libraries.

hzxuzhonghu:
ok, will do

pkg/controller.v1/common/job_controller.go
hzxuzhonghu: One way is to update `NewJobController`, and add a new argument to indicate whether to use volcano or kube-batch. Then it will require all the operators to change. I think it is not worth doing that.

Jeffwan:
looks good to me

hzxuzhonghu:
Thank you all

Issue #48, opened 2019-10-15
Delete direct dependency of kubernetes


liubog2008:
Now go.mod has dependency of kubernetes v1.11.2.
It will be hard to import job_controller/pkg/api because of compatibility with new version client-go

liubog2008:
```
$ go mod why k8s.io/client-go/util/integer 
go: finding k8s.io/api latest
go: finding k8s.io/apimachinery latest
# k8s.io/client-go/util/integer
xxx/cmd
github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned
github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned/typed/kubeflow/v1alpha2
github.com/kubeflow/mpi-operator/pkg/apis/kubeflow/v1alpha2
github.com/kubeflow/common/job_controller/api/v1
k8s.io/kubernetes/pkg/controller
k8s.io/client-go/util/integer
```

gaocegege:
We should to remove the dep here https://github.com/kubeflow/common/blob/master/job_controller/api/v1/controller.go#L83

Then we do not need "k8s.io/kubernetes/pkg/controller"

liubog2008:
There are two way to resolve this problem

1. Fork `controller.ControllerExpectationsInterface` as third_party code
2. Move `JobController` out of package `api/v1` (why it is in v1 package ==)

gaocegege:
I have no idea why we place it in api.

@richardsliu @jian-he @merlintang @terrytangyuan Do you have idea about it?

mouradmourafiq:
Any update on this issue, it's breaking dependency on all operators' api packages.

Jeffwan:
> There are two way to resolve this problem
> 
> 1. Fork `controller.ControllerExpectationsInterface` as third_party code

em. this will work but probably not the elegant way, I notice there's an issue to suggest to add expectations in client-go. I reopen the issue and will spend some time on it later. https://github.com/kubernetes/client-go/issues/332

> 2. Move `JobController` out of package `api/v1` (why it is in v1 package ==)

I doubt this will help resolve the dependency issue if you use go mod.? Correct me if I am wrong. Even you import `github.com/kubeflow/common/job_controller/api/v1`, dependencies in other packages will be fetched to dependency graph. But I agree it would be good to move type JobController out of api package. It should be part of the  `https://github.com/kubeflow/common/blob/master/job_controller/job_controller.go`

[...]

PR #72, opened 2020-04-15
Move Pods/Services control interface to separate folder

Jeffwan:
Currently, pods/services implementation mixed with jobController together under `common` folder, this is a little messy. 

The reason we want to have separate folder (package name) for it are
1. PodControlInterface is folked from [controller_util.go](https://github.com/kubernetes/kubernetes/blob/release-1.15/pkg/controller/controller_utils.go) and we also implement similar codes for service. They are logically separate. For long term, once https://github.com/kubernetes/client-go/issues/332 is merged, we could remove the folder entirely.

2.  Since `PodControlInterface` and `ServiceControlInterface` are internal code base for common project, there's no need for users to import them at all. 

3. Trying to make some changes inside `control` to solve #48. It would be better to file this PR first and make changes against new folder, that would be easier for review. Otherwise, it's a new file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	k8s.io/client-go v0.16.9-beta.0
 	k8s.io/code-generator v0.15.10
 	k8s.io/kube-openapi v0.0.0-20190228160746-b3a7cee44a30
-	k8s.io/kubernetes v1.18.0
+	k8s.io/kubernetes v1.16.2
 	volcano.sh/volcano v0.4.0
 )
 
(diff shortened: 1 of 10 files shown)


terrytangyuan:
@Jeffwan Wwe also need to update the import statement in xgboost-operator

Jeffwan:
@terrytangyuan What's the statement? Do you need me to help update common there?

terrytangyuan:
@Jeffwan Never mind. It looks like there is no public interface change yet so I am only updating the version of common in https://github.com/kubeflow/xgboost-operator/pull/47.

Jeffwan:
> @Jeffwan Never mind. It looks like there is no public interface change yet so I am only updating the version of common in [kubeflow/xgboost-operator#47](https://github.com/kubeflow/xgboost-operator/pull/47).

Leave you comment on the PR.

Issue #71, opened 2020-04-14
Difference between kubeflow/common and kubeflow/tf-operator/pkg/common


Jeffwan:
Comparing `ControllerInterface` between two different projects, tf-operator is pretty clean. `kubeflow/common` adds extra methods to create Job/Pod/Services. 

https://github.com/kubeflow/common/blob/fda69c7bdd3870ab5981538542c2e6b01fe076be/pkg/apis/common/v1/interface.go#L31-L56

Besides that, we extracted a few interfaces for high level abstraction which makes sense. https://github.com/kubeflow/common/blob/fda69c7bdd3870ab5981538542c2e6b01fe076be/pkg/apis/common/v1/interface.go#L59-L72

JobController side, tf-operator has `PodControl` and `ServiceControl` which are used to operate k8s objects. However, in https://github.com/kubeflow/common/pull/36, since create/delete pods/services interface have been exposed, community determine to remove `PodControl` and `ServiceControl` because them are kubernetes controller internal codes, and there's also concern on the size of the dependency (need kubernetes code base). 

Currently, it's a little bit weird. 

1. We can not fully get rid of `kubernetes/pkg/controller`, Expectation is still using them here. https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/job_controller.go#L100 . This is not part of `client-go` yet. See issue https://github.com/kubernetes/client-go/issues/332

I think we used `dep` to manage dependencies in the past and I've migrated common to go module compatible. Ideally, we will move all operator to go modules pretty soon. Size of the dependencies is not that painful. 

2. Files https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/pod_control.go and https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/service_control.go are not being used in `kubeflow/common`. 

I think we only one of following three. The status in `kubeflow/common` is 

- Create Pods/Service Interface  - exist and being used 
- controller. ServiceControlInterface - deleted
- copied ServiceControlInterface - still there and but not being used. Only tests are using it.

3. Implementation from different operators for following interface are very close. I doubt we need different controller to have similar logic there. Currently, only xgboost-operator is built on top of this operator, we have plan to migrate tf, pytorch, mxnet to follow kubeflow/common fashion.  trying to see any other operators are built on top of common, if there's a need to customize logic of `CreateService`, etc. If that's all same, it's better to implement it inside JobController. 

[...]

PR #73, opened 2020-04-15
Add PodControlInterface and PodControllerRefManager

Jeffwan:
This PR aims to solve following issue and it also help to remove more direct dependency on Kubernetes.  Resolve part of #48 #71 

### Background
Current implementation uses `PodControlInterface` and `PodControllerRefManager` from `k8s.io/kubernetes/pkg/controller` and self-implemented `ServiceControlInterface` and `ServiceControllerRefManager` which bring some problems.

1. PodControlInterface/PodControllerRefManager may change along with k8s version upgrade, while we don't make changes to Service. It maybe out of sync.
2. Half and half doesn't help us remove Kubernetes direct dependency.
3. It’s not that straightforward for maintainers and contributors. The first time I look at the change, I notice we only have `service_ref_manager` which is kind of unclear until I find #36 

### Solution
Consider the situation that Kubernetes upstream doesn't have implementation of `ServiceControlInterface` and `ServiceControllerRefManager`, I would suggest we just folk `PodControlInterface` and `PodControllerRefManager` locally.

This help remove dependency a lot and we can also make sure pods/service implementation can be updated at the same time. 

In addition, I will make some efforts to this issue. https://github.com/kubernetes/client-go/issues/332. I will try to submit a PR and see feedbacks. If change can be accepted, then we don't need to maintain these auxiliary codes on our side. 

/cc @gaocegege @terrytangyuan 

Need @gaocegege 's review on code folk since he originally worked on this.


pkg/controller.v1/control/controller_ref_manager.go
```diff
@@ -0,0 +1,377 @@
+// Copyright 2019 The Kubeflow Authors
```
Jeffwan, line 1: If we think it's fine to sync pod_ref_manager, it makes sense to rename to `controller_ref_manager.go`.

pkg/controller.v1/control/service_control.go
```diff
@@ -16,6 +16,7 @@ package control
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
```
gaocegege, line 19: Please move this to the next group

Jeffwan: Yeah, I move to following block.

Jeffwan:
@gaocegege Address the feedbacks, please have another look

gaocegege:
> LGTM but could you add some notes in the forked code so we can always look back what we did and where the code was forked from (with version/link).

+1 for this suggestion.

Jeffwan:
@gaocegege @terrytangyuan 

Thanks for the review. That's great suggestion. I will add a follow up CONTRIBUTING.md doc and include these technical details.

gaocegege:
OK, then I think we can merge this one.

PR #77, opened 2020-04-23
Add proposal for Prometheus metrics coverage

terrytangyuan:
This provides a detailed outline of the Prometheus metrics we plan to coverage in common operator. Related issue: https://github.com/kubeflow/common/issues/22.

Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>

### DIFF
--- a/docs/prometheus-metrics.md
+++ b/docs/prometheus-metrics.md
@@ -7,30 +7,30 @@ The goal is to report generic metrics (e.g. metrics related to pods/jobs/service
 * It is easier for users of Kubeflow distributed training operators to monitor operator performance and behaviors using consistent set of metrics for different distributed training operators.
 
 This document outlines the list of Prometheus metrics we plan to cover in `JobController`. We follow the metric naming convention
-outlined [here](https://prometheus.io/docs/practices/naming/).
+outlined [here](https://prometheus.io/docs/practices/naming/) and the metric types supported by Prometheus [here](https://prometheus.io/docs/concepts/metric_types/).
 
 ## Pod Metrics
 
 The following metrics related to the lifecycle of pods will be reported:
 
-| Metric Name | Description |
-| ------------ | ------- |
-| created_pods_total | The total number of created pods |
-| restarted_pods_total | The total number of restarted pods |
-| deleted_pods_total | The total number of deleted pods |
-| failed_pods_total | The total number of failed pods |
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| created_pods_total | Counter | The total number of created pods |
+| restarted_pods_total | Counter | The total number of restarted pods |
+| deleted_pods_total | Counter | The total number of deleted pods |
+| failed_pods_total | Counter | The total number of failed pods |
 
 The following metrics will be reported on each pod:
 
-| Metric Name | Description |
-| ------------ | ------- |
-| container_cpu_usage_seconds_total | CPU usage |
-| container_accelerator_memory_used_bytes | GPU usage |
-| container_memory_usage_bytes | Memory usage |
-| container_network_transmit_bytes_total | Network usage |
-| container_fs_write_seconds_total | I/O usage |
-| up | Keep-Alive check (maintained by Prometheus on its own with its `up` metric detailed in the documentation [here](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series))) |
-| common_operator_is_leader | Whether this client is the leader of this common operator client set |
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| container_cpu_usage_seconds_total | Counter | Cumulative cpu time consumed in seconds |
+| container_accelerator_memory_used_bytes | Gauge | Total accelerator memory allocated |
+| container_memory_usage_bytes | Gauge | Current memory usage in bytes, including all memory regardless of when it was accessed |
+| container_network_transmit_bytes_total | Counter | Cumulative count of bytes transmitted |
+| container_fs_usage_bytes | Gauge | Number of bytes that are consumed by the container on this filesystem |
+| up | Gauge | Keep-Alive check (maintained by Prometheus on its own with its `up` metric detailed in the documentation [here](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series))) |
+| common_operator_is_leader | Gauge | Whether this client is the leader of this common operator client set |
 
 Note that some of the above metrics are derived from [cAdvisor](https://github.com/google/cadvisor) kubelet
 integration which reports to Prometheus through our prometheus-operator installation.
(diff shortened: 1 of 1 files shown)


docs/prometheus-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+* Other operators built on top of it will automatically report Prometheus metrics without additional efforts;
+* It is easier for users of Kubeflow distributed training operators to monitor operator performance and behaviors using consistent set of metrics for different distributed training operators.
+
+This document outlines the list of Prometheus metrics we plan to cover in `JobController`.
+
+## Pod Metrics
+
+The following metrics related to the lifecycle of pods will be reported:
+
+* The total number of created pods
```
gaocegege, line 15: I think we can list the name of the metrics here.

Jeffwan: Yeah. That would be great. We can use them directly in implementation.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+
+The following metrics will be reported on each pod:
+
+* CPU usage
+* GPU usage
+* Memory usage
+* Network usage
+* I/O usage
+* Keep-Alive check
+* Is-leader check
```
Jeffwan, line 28: the leader means training job leader?

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Job Metrics
+
+The following metrics related to the lifecycle of jobs will be reported:
```
Jeffwan, line 32: I think we also need duration for some job state transformation. Like created -> started, started to finished, etc.

ywskycn: +1, duration among different states is quite useful when debugging in production.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Pod Metrics
+
+The following metrics related to the lifecycle of pods will be reported:
+
+* The total number of created pods
+* The total number of restarted pods
+* The total number of deleted pods
+* The total number of failed pods
+
```
merlintang, line 19: How about the metric of job itself ? like the job start time, and end time?

ywskycn: this can be covered by @Jeffwan 's suggestions.

terrytangyuan:
Thanks everyone for the comments! I've converted the lists to tables which include the metric name, type, and description. I also added a few additional metrics as suggested. Hopefully it's much clearer now. Please take another look.

docs/prometheus-metrics.md
terrytangyuan: Thanks. Added.

docs/prometheus-metrics.md
terrytangyuan: I renamed it and added description. Please let me know if it's still unclear.

docs/prometheus-metrics.md
terrytangyuan: Good suggestion. Added.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,82 @@
...
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| container_cpu_usage_seconds_total | Counter | Cumulative cpu time consumed in seconds |
+| container_accelerator_memory_used_bytes | Gauge | Total accelerator memory allocated |
+| container_memory_usage_bytes | Gauge | Current memory usage in bytes, including all memory regardless of when it was accessed |
+| container_network_transmit_bytes_total | Counter | Cumulative count of bytes transmitted |
+| container_fs_usage_bytes | Gauge | Number of bytes that are consumed by the container on this filesystem |
+| up | Gauge | Keep-Alive check (maintained by Prometheus on its own with its `up` metric detailed in the documentation [here](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series))) |
+| common_operator_is_leader | Gauge | Whether this client is the leader of this common operator client set |
```
johnugeorge, line 33: do you mean, the elected leader when redundant operator pods created ? How do you define this from Jobcontroller library perspective?

terrytangyuan: Yes, this might be tricky as it involves moving the leader election related code to kubeflow/common. We can certainly do it at other operators if it doesn't fit in JobController.

johnugeorge: Yes. Currently, this is not part of kubeflow/common. we could keep this out of scope for current time being.

terrytangyuan: Sounds good. I removed it from the table.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+| ----------- | ------------| ----------- |
+| container_cpu_usage_seconds_total | Counter | Cumulative cpu time consumed in seconds |
+| container_accelerator_memory_used_bytes | Gauge | Total accelerator memory allocated |
+| container_memory_usage_bytes | Gauge | Current memory usage in bytes, including all memory regardless of when it was accessed |
+| container_network_transmit_bytes_total | Counter | Cumulative count of bytes transmitted |
+| container_fs_usage_bytes | Gauge | Number of bytes that are consumed by the container on this filesystem |
+| up | Gauge | Keep-Alive check (maintained by Prometheus on its own with its `up` metric detailed in the documentation [here](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series))) |
+
+Note that some of the above metrics are derived from [cAdvisor](https://github.com/google/cadvisor) kubelet
+integration which reports to Prometheus through our prometheus-operator installation.
```
Jeffwan, line 35: Want to make sure the scope. This is outside operator. By default cadvisor expose the metrics and user can use these by their own.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+| restarted_jobs_total | Counter | The total number of restarted jobs |
+| pending_jobs_total | Counter | The total number of pending jobs |
+| failed_jobs_total | Counter | The total number of failed jobs |
+| running_jobs_total | Counter | The total number of running jobs |
+
+The following metrics related to the duration among various job phases will be reported:
+
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| from_created_to_completed_job_duration_seconds_total | Counter | The duration between job created and job completed in seconds |
```
Jeffwan, line 55: Minor: I am thinking if we should change to `job_duration_from_created_to_complated_seconds_total`.  Another thing is seems it would be good to use `complete` `deleted` as labels, but duration requires two and it would be a little bit hard to query. I think adding labels into metrics to distinguish them makes sense.

Jeffwan:
Beside above minor comments, it looks good to me. Wait to see if someone else has the feedback

docs/prometheus-metrics.md
terrytangyuan: Yes, but I think it's good to document this here so we know that we don't need to report these metrics by ourselves.

docs/prometheus-metrics.md
terrytangyuan: I am following the naming practice outlined here: https://prometheus.io/docs/practices/naming/. I prefer the current naming without label as it's more intuitive but we can certainly revisit/revise later.

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| created_jobs_total | Counter | The total number of created jobs |
+| deleted_jobs_total | Counter | The total number of deleted jobs |
+| completed_jobs_total | Counter | The total number of completed jobs |
+| restarted_jobs_total | Counter | The total number of restarted jobs |
+| pending_jobs_total | Counter | The total number of pending jobs |
+| failed_jobs_total | Counter | The total number of failed jobs |
+| running_jobs_total | Counter | The total number of running jobs |
```
yeya24, line 49: I am curious about this one. Why `running job` is a counter? Counter is usually used with `rate`, but as for `running job`, we want to know the current jobs in the running status. So I think it is more appropriate to use `Gauge` type, WDYT?

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+
+The following metrics related to the duration among various job phases will be reported:
+
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| from_created_to_completed_job_duration_seconds_total | Counter | The duration between job created and job completed in seconds |
+| from_completed_to_deleted_job_duration_seconds_total | Counter | The duration between job completed and job deleted in seconds |
+| from_failed_to_restarted_job_duration_seconds_total | Counter | The duration between job failed and job restarted in seconds |
+| from_pending_to_running_job_duration_seconds_total | Counter | The duration between job pending and job running in seconds |
```
yeya24, line 58: Why these metrics are `Counter`? Maybe for duration metrics, it is better to use `histogram`?

docs/prometheus-metrics.md
yeya24: The same for `pending jobs`

terrytangyuan: Good catch. Thanks. Fixed!

docs/prometheus-metrics.md
terrytangyuan: Great suggestion. Updated.

terrytangyuan:
@yeya24 Thanks! Great suggestions. I have updated the metric types in the doc.

PTAL @gaocegege @johnugeorge

docs/prometheus-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+The following metrics related to the lifecycle of jobs will be reported:
+
+| Metric Name | Metric Type | Description |
+| ----------- | ------------| ----------- |
+| created_jobs_total | Counter | The total number of created jobs |
+| deleted_jobs_total | Counter | The total number of deleted jobs |
+| completed_jobs_total | Counter | The total number of completed jobs |
+| restarted_jobs_total | Counter | The total number of restarted jobs |
+| pending_jobs_total | Gauge | The total number of pending jobs |
+| failed_jobs_total | Counter | The total number of failed jobs |
```
yeya24, line 48: @terrytangyuan  Forgot to mention this one. Do you think it is more appropriate to make this `Gauge` as well?  Do you want to represent the history failures or the current failed jobs?

Can we list the metrics label in this doc as well? This is important and useful, too. Like we can combine `pending jobs` `running jobs` and `failed jobs` into one metric job_status{status="pending/failed/running"}, WDYT?

terrytangyuan: Let's keep it as it is for now so that the metrics are consistent for metrics with past tense v.s. metrics with present continuous tense. Currently there are no labels yet as it's hard to differentiate metrics with two different tenses and choose different metric types for those metrics.

PR #85, opened 2020-05-13
Simplify interfaces and remove useless interfaces

Jeffwan:
The motivation of this PR is to reduce the efforts to build a training job operators. 

1. These are all common CRUD operations and there's no big difference between job operators. We don't have to expose to external for implementation. Instead, this should be inside common libraries or inbuilt controller methods user should not take care of. 

```
GetPodsForJob
GetServiceForJob
CreateService
DeleteService
CreatePod
DeletePod
```

As we bring back control and expectation packages, we have enough utils to implement these methods easily inside `JobController`. 

2. `GetDefaultContainerPortNumber()` is an interface never been used, we reply on `GetDefaultContainerName()` and `GetDefaultContainerPortName ()` to find the port, this is expose to user to make the decision. I think it was brought into interface but not well designed.. We remove it for now.

### DIFF
--- a/pkg/controller.v1/common/pod.go
+++ b/pkg/controller.v1/common/pod.go
@@ -49,7 +49,7 @@ const (
 	// podTemplateSchedulerNameReason is the warning reason when other scheduler name is set
 	// in pod templates with gang-scheduling enabled
 	podTemplateSchedulerNameReason = "SettedPodTemplateSchedulerName"
-
+	// gangSchedulingPodGroupAnnotation is the annotation key used by batch schedulers
 	gangSchedulingPodGroupAnnotation = "scheduling.k8s.io/group-name"
 )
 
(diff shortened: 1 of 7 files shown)


Jeffwan:
Just finish two week's oncall and come back to work. Hope to make common stabilize this week and we can start operator migration

merlintang:
yes, does this impact the xgboost operator?

Jeffwan:
```
pkg/controller.v1/common/service.go:37:2: var `succeededServiceCreationCount` is unused (unused)
pkg/controller.v1/common/service.go:41:2: var `failedServiceCreationCount` is unused (unused)
The command "golangci-lint run --config=linter_config.yaml ./..." exited with 1.
```

This is a rebase failure since metrics was on the old version. Add them back in latest code snippet.

Jeffwan:
> yes, does this impact the xgboost operator?

Right, xgboost operator needs to make some changes to use this version. I will cut a release later and help xgboost migration.

pkg/controller.v1/common/pod.go
```diff
@@ -437,10 +477,15 @@ func (jc *JobController) createNewPod(job interface{}, rt, index string, spec *a
 		} else {
 			podTemplate.Spec.SchedulerName = gangSchedulerName
 		}
+
+		if podTemplate.Annotations == nil {
+			podTemplate.Annotations = map[string]string{}
+		}
+		podTemplate.Annotations[gangSchedulingPodGroupAnnotation] = metaObject.GetName()
```
merlintang, line 484: i am not sure we need to have gangscheduling here, since we might do not need gangscheduling.

Jeffwan: @merlintang 

Yeah. it's up to the controller 

You can disable that when you create job controller
https://github.com/kubeflow/xgboost-operator/blob/master/pkg/controller/xgboostjob/xgboostjob_controller.go#L131

Here, I have a minor improvement. Instead of create annotation for all operators, I check configuration first and only create this annotation if operator enable it.

PR #87, opened 2020-05-14
Bump Kubernetes dependency to 1.16.9

### LINKED ISSUE #86: Bump Kubernetes dependency to 1.16
I've seen this issue when I use common in other operators. The problem is volcano changes brings Kubernetes 1.16 and it has direct dependency on 1.16. At the same time, we still replace k8s packages to 1.15.10. 

Build will work but if we determine to use 1.16 in other projects, CRD client.go will have incompatibility issue like beflow. 

```
../../../../pkg/mod/github.com/kubeflow/common@v0.3.0/pkg/util/k8sutil/client.go:50:32: undefined: serializer.DirectCodecFactory
```

Let's bump version to 1.16 and make sure they are consistent.

Jeffwan:
Resolve #86 

client.go is still using 1.15 dependency. Since volcano brings 1.16 already, if end controller use 1.16, it will have problem like below. This PR make them consistent.

```
../../../../pkg/mod/github.com/kubeflow/common@v0.3.0/pkg/util/k8sutil/client.go:50:32: undefined: serializer.DirectCodecFactory
```

1. Bump Kubernetes dependency to 1.16.9
2. Update CRD client.go to 1.16 compatible
3. Update expectation package to 1.16 compatible


Jeffwan:
```
--- ./hack/../pkg/apis/common/v1/openapi_generated.go	2020-05-14 06:27:08.721327529 +0000
+++ ./hack/../_tmp/pkg/apis/common/v1/openapi_generated.go	2020-05-14 06:25:01.000000000 +0000
```

Test failed because of `openapi_generated.go`. I regenerate for 1.16 version.

PR #100, opened 2020-08-04
minor errors in pod.go

xfate123:
just some minor errors I found when I studied the code

### DIFF
--- a/pkg/controller.v1/common/pod.go
+++ b/pkg/controller.v1/common/pod.go
@@ -24,7 +24,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -86,7 +86,7 @@ func (jc *JobController) AddPod(obj interface{}) {
 		logger := commonutil.LoggerForPod(pod, jc.Controller.GetAPIGroupVersionKind().Kind)
 
 		if job == nil {
-			if pod.Labels[apiv1.GroupNameLabel] == jc.Controller.GetGroupNameLabelValue() {
+			if pod.Labels[apiv1.GroupNameLabel] != jc.Controller.GetGroupNameLabelValue() {
 				logger.Info("This pod's job does not exist")
 			}
 			return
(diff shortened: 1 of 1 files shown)


pkg/controller.v1/common/pod.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
terrytangyuan, line 27: This change is unnecessary

pkg/controller.v1/common/pod.go
```diff
@@ -392,9 +392,9 @@ func (jc *JobController) ReconcilePods(
 			}
 			// Check if the pod is retryable.
 			if spec.RestartPolicy == apiv1.RestartPolicyExitCode {
-				if pod.Status.Phase == v1.PodFailed && trainutil.IsRetryableExitCode(exitCode) {
+				if pod.Status.Phase == v1.PodFailed && !trainutil.IsRetryableExitCode(exitCode) {
```
terrytangyuan, line 395: I think this was intended but we just need to improve the log message

pkg/controller.v1/common/pod.go
```diff
@@ -86,7 +86,7 @@ func (jc *JobController) AddPod(obj interface{}) {
 		logger := commonutil.LoggerForPod(pod, jc.Controller.GetAPIGroupVersionKind().Kind)
 
 		if job == nil {
-			if pod.Labels[apiv1.GroupNameLabel] == jc.Controller.GetGroupNameLabelValue() {
+			if pod.Labels[apiv1.GroupNameLabel] != jc.Controller.GetGroupNameLabelValue() {
```
merlintang, line 89: good catch up

pkg/controller.v1/common/pod.go
gaocegege: It is created by gofmt I think, maybe we can keep it.

pkg/controller.v1/common/pod.go
Jeffwan: The original logic here is to provide flexibility to define your own retryable exit code. This should not be changed. 

Since kubernetes doesn't have restart method, That's the reason we `delete` the pod, and wait for next `reconcile` loop to create a new one. It's asynchronous so `restart` here may be confused to users.

pkg/controller.v1/common/pod.go
Jeffwan: @xfate123 

Please check logics here. https://github.com/kubeflow/common/blob/ac0c6e151f871bd7ed5673648b021f8351b5c663/pkg/controller.v1/common/job_controller.go#L321-L336

If we can not find the job, we will end reconcile loop and directly `return`. The only thing matters is if we want to have some meaningful log. 

There're several reason `job == nil` here

1. Kind doesn't match
2. GroupNameLabel doesn't match
3. Job doesn't exist. In this case, the pod is an orphan pod.

the 3rd one is the only case we want to persist the log. Does it make sense?

xfate123: @Jeffwan 
Thank you so much Jeff. Much clearer after your patient explanation.
But I am still a little confused
I think the three cases should be:
1.Kind doesn't match
2.Name unmatched
3. Name matched but UID unmatched
My understanding is that third case cannot prove this pod is an orphan pod. It can only prove that the controllerRef point s to a different job.
Thank you again. Really appreciate your help
Thank you

Jeffwan: You are right. I didn't follow logic in `resolveControllerRef`. UID mismatch is one of the cases.  We check `GroupNameLabel` again in caller side which I think is unnecessary. We use fixed value `kubeflow.org` for most of the operators.  

> My understanding is that third case cannot prove this pod is an orphan pod. It can only prove that the controllerRef point s to a different job.

I may not explain this clearly. you are right.  It could be a pod without Job ref or a pod point to different job (UID mismatch).

xfate123: @Jeffwan  Thank you so much for your clear explanation

Issue #104, opened 2020-12-29
CleanPodPolicyRunning should also clean Pending pods


mkkb473:
https://github.com/kubeflow/common/blob/31a6ec10d140a3c8cb12a863dcc1363f46ad9998/pkg/controller.v1/common/job.go#L31
orphan Pending pod will become orphan Running pod once schedulable.

PR #105, opened 2020-12-29
CleanPodPolicyRunning cleans Pending phase pod

mkkb473:
#104

### DIFF
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -28,9 +28,9 @@ func (jc *JobController) DeletePodsAndServices(runPolicy *apiv1.RunPolicy, job i
 	}
 
 	for _, pod := range pods {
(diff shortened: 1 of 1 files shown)


ChanYiLin:
/lgtm
Thanks for your contribution !

gaocegege (commented):
/lgtm

/assign @Jeffwan @johnugeorge @andreyvelich

ChanYiLin:
just some wording issue in the comment

//Note that pending pod will ~~turn to be~~ turn into (or turn) running once schedulable,
// not cleaning it may leave ~~there~~ orphan running pod"s" in the ~~fure~~ future,
// we should ~~treet~~  treat it equivalent to running phase here.

Jeffwan:
Nice catch. This would be helpful in some cases. 

/lgtm
/approve

PR #120, opened 2021-03-27
update volcano.sh/volcano to volcano.sh/apis

Thor-wl:
Signed-off-by: Thor-wl <1187526662@qq.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	k8s.io/client-go v0.16.9
 	k8s.io/code-generator v0.16.9
 	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf
-	volcano.sh/volcano v0.4.0
+	volcano.sh/apis v1.16.15
 )
 
 replace (
--- a/pkg/controller.v1/common/job_controller.go
+++ b/pkg/controller.v1/common/job_controller.go
@@ -26,8 +26,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
-	"volcano.sh/volcano/pkg/apis/scheduling/v1beta1"
-	volcanoclient "volcano.sh/volcano/pkg/client/clientset/versioned"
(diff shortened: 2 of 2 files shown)


terrytangyuan (approved):
PTAL 

/lgtm

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	k8s.io/client-go v0.16.9
 	k8s.io/code-generator v0.16.9
 	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf
-	volcano.sh/volcano v0.4.0
+	volcano.sh/apis v1.16.15
```
shinytang6, line 20: l think we also need to update the gomod dependency?

pkg/controller.v1/common/job_controller.go
```diff
@@ -26,8 +26,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
-	"volcano.sh/volcano/pkg/apis/scheduling/v1beta1"
-	volcanoclient "volcano.sh/volcano/pkg/client/clientset/versioned"
+	"volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	volcanoclient "volcano.sh/apis/pkg/client/clientset/versioned"
```
shinytang6, line 30: There is another place using volcano.sh/volcano directly.

link: [https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/job.go#L18](https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/job.go#L18)

go.mod
Thor-wl: Yes, so i close the pr and submit a new one

terrytangyuan: Why not updating this PR directly?

Thor-wl: Is there any buttons or labels available to restart the CI test?

terrytangyuan: When you push a new commit it will trigger new builds.

Thor-wl: OK, Thanks.

PR #123, opened 2021-03-29
update Volcano CRD API to v1.16

Thor-wl:
Signed-off-by: Thor-wl <13164644535@163.com>


Thor-wl:
I've got the root reason for CI failed. Before executing `go build`, `go get volcano.sh/apis@v1.16.15` works and pull the code from volcano.sh/apis. Indeed, it pulls the right version. But what confused me is that all `zz_generated.xx,go` files under each CRD diretory were not pulled. Do you have met this problem? Any advice?

Thor-wl:
Well, what were pulled to local is not from the repository(not the up-to-date code) but looks like a old version. There seems to be some cache code. Do you have some ideas about that?

shinytang6:
> Well, what were pulled to local is not from the repository(not the up-to-date code) but looks like a old version. There seems to be some cache code. Do you have some ideas about that?

let me have a try :)

Jeffwan:
> Well, what were pulled to local is not from the repository(not the up-to-date code) but looks like a old version. There seems to be some cache code. Do you have some ideas about that?

After checking logs, I notice volcano api uses 1.19 changes which already has context support in client operations. However, in kubeflow/common, we replace dependencies to 1.16 which doesn't have it. That's the reason build failed. 

Let me file a PR to bump the version and you can rebase the changes, then change PR title to "Use Volca CRD API"  (Remove version information)

Thor-wl:
> > Well, what were pulled to local is not from the repository(not the up-to-date code) but looks like a old version. There seems to be some cache code. Do you have some ideas about that?
> 
> After checking logs, I notice volcano api uses 1.19 changes which already has context support in client operations. However, in kubeflow/common, we replace dependencies to 1.16 which doesn't have it. That's the reason build failed.
> 
> Let me file a PR to bump the version and you can rebase the changes, then change PR title to "Use Volca CRD API" (Remove version information)

I've taken some tests about it. It's caused by the reuse of tag v1.16.15. Namely, i've had taged v1.16.15 ,deleted it, made some changes and taged again. But the go repository cached the original version and will not be refreshed again. So we can only get the old version from cache. I'll make a new tag.

Jeffwan:
Please have a check on https://github.com/kubeflow/common/pull/126. Once it's merged, this PR should be unblocked (need minor changes)

Thor-wl:
I've update the tag and it's ok now.

go.mod
```diff
@@ -10,12 +10,12 @@ require (
...
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.4.0
-	k8s.io/api v0.16.9
-	k8s.io/apimachinery v0.16.9
-	k8s.io/client-go v0.16.9
-	k8s.io/code-generator v0.16.9
+	k8s.io/api v0.16.15
+	k8s.io/apimachinery v0.16.15
+	k8s.io/client-go v0.16.15
+	k8s.io/code-generator v0.18.3
```
Jeffwan, line 16: code-generator version should match with other k8s versions

Jeffwan: this is a minor issue.

Thor-wl: > this is a minor issue.

Oh, yes. Sorry for ignoring the mismatch. It seems that it is modified automatically whe executing `go mod vendor`. I'll submit another pr for it.

Jeffwan: never mind. probably due to some dependency issue. We can clean up dependencies together

shinytang6:
great, LGTM :)
/approve

gaocegege:
/lgtm

PR #126, opened 2021-03-30
Bump k8s dependency to 1.19 and add context support

Jeffwan:
1. Upgrade k8s dependency to 1.19
2. 1.18 client-go add context and extra options. Adapt to the new changes.
3. Upgrade golang to 1.15 to match golang used in 1.19


gaocegege:
I think the main purpose of https://github.com/kubeflow/common/pull/123 is to avoid the direct dependency to k8s. Then I am wondering if we should upgrade the version to 1.19.

Do you want to add context first, then we can eliminate the k8s import packages?

Thor-wl:
I'm both OK for avoiding the direct depend dependency to k8s first or upgrade to k8s to v1.19 first. You can make a decision @Jeffwan @gaocegege

Jeffwan:
> I think the main purpose of #123 is to avoid the direct dependency to k8s. Then I am wondering if we should upgrade the version to 1.19.
> 
> Do you want to add context first, then we can eliminate the k8s import packages?

@gaocegege I think your concern is reasonable. community is releasing 1.21. It's better to keep < 3 version changes with latest version and we will update to higher version sometime soon. 

1.18 brings in context changes, I am ok to bump to 1.18. The major reason I choose 1.19 is because volcano/api now is using 1.19 https://github.com/volcano-sh/apis/blob/cfacea273e7da6b53170cc00ece9b2caf019316f/go.mod#L6-L10

Thor-wl:
> > I think the main purpose of #123 is to avoid the direct dependency to k8s. Then I am wondering if we should upgrade the version to 1.19.
> > Do you want to add context first, then we can eliminate the k8s import packages?
> 
> @gaocegege I think your concern is reasonable. community is releasing 1.21. It's better to keep < 3 version changes with latest version and we will update to higher version sometime soon.
> 
> 1.18 brings in context changes, I am ok to bump to 1.18. The major reason I choose 1.19 is because volcano/api now is using 1.19 https://github.com/volcano-sh/apis/blob/cfacea273e7da6b53170cc00ece9b2caf019316f/go.mod#L6-L10
> 
> Another possible solution is to ask volcano/api to downgrade to 1.17 and we don't need client-go related changes. But we will still need to upgrade dependencies later.

It's OK for me to provide volcano.sh/apis release-1.17 to match k8s.io v1.17. Look forward to your decision.

Jeffwan:
`volcano.sh/apis v1.2.0-k8s1.16.15` this makes sense. I think @gaocegege will like it. :D 
In this case, we will not bump k8s version at this moment, this PR can be left for future usage.

gaocegege:
SGTM! I think we will update to a higher version soon.

Thor-wl:
> `volcano.sh/apis v1.2.0-k8s1.16.15` this makes sense. I think @gaocegege will like it. :D
> In this case, we will not bump k8s version at this moment, this PR can be left for future usage.

> SGTM! I think we will update to a higher version soon.

@Jeffwan @gaocegege So what's the version shall i provide?

Jeffwan:
/hold cancel 

This unblocks user's request @chxk in https://github.com/kubeflow/common/issues/132

terrytangyuan (approved):
LGTM. Others PTAL.

andreyvelich (approved):
Thank you for doing this @Jeffwan!
/lgtm

PR #130, opened 2021-04-06
feat(common): support delay pods creation

### LINKED ISSUE #129: [feature] Support delay pods creating when EnableGangScheduling is true
As described in [https://github.com/kubeflow/tf-operator/issues/1264](https://github.com/kubeflow/tf-operator/issues/1264) and [https://github.com/volcano-sh/volcano/issues/1387](https://github.com/volcano-sh/volcano/issues/1387)

we expect that if `EnableGangScheduling` is true, the pods controlled by job operator should be delay creating util the podgroup status is `Inqueue` or `Running`.

/cc @Thor-wl @jiangkaihua

shinytang6:
resolves #129

### DIFF
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -307,9 +307,15 @@ func (jc *JobController) ReconcileJobs(
 				MinResources:      minResources,
 			}
 
-			_, err := jc.SyncPodGroup(metaObject, pgSpec)
+			pg, err := jc.SyncPodGroup(metaObject, pgSpec)
 			if err != nil {
 				log.Warnf("Sync PodGroup %v: %v", jobKey, err)
+				return nil
+			}
+
+			if pg == nil || pg.Status.Phase == "" || pg.Status.Phase == v1beta1.PodGroupPending {
+				log.Warnf("PodGroup %v unschedulable", jobKey)
+				return nil
 			}
 		}
 
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -307,15 +307,25 @@ func (jc *JobController) ReconcileJobs(
 				MinResources:      minResources,
 			}
 
+			syncReplicas := true
 			pg, err := jc.SyncPodGroup(metaObject, pgSpec)
 			if err != nil {
 				log.Warnf("Sync PodGroup %v: %v", jobKey, err)
-				return nil
+				syncReplicas = false
 			}
 
+			// Delay pods creation until podgroup status is inqueue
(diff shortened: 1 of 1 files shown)


pkg/controller.v1/common/job.go
```diff
@@ -307,9 +307,15 @@ func (jc *JobController) ReconcileJobs(
...
-			_, err := jc.SyncPodGroup(metaObject, pgSpec)
+			pg, err := jc.SyncPodGroup(metaObject, pgSpec)
 			if err != nil {
 				log.Warnf("Sync PodGroup %v: %v", jobKey, err)
+				return nil
+			}
+
+			if pg == nil || pg.Status.Phase == "" || pg.Status.Phase == v1beta1.PodGroupPending {
+				log.Warnf("PodGroup %v unschedulable", jobKey)
+				return nil
```
gaocegege, line 318: Did we watch PodGroup?

I am not sure if there is a new reconciliation loop for the status change.

shinytang6: seems not, maybe we need to update the job status here to trigger a new reconciliation loop, l will give it a try.

pkg/controller.v1/common/job.go
```diff
@@ -307,9 +307,15 @@ func (jc *JobController) ReconcileJobs(
...
 			}
 
-			_, err := jc.SyncPodGroup(metaObject, pgSpec)
+			pg, err := jc.SyncPodGroup(metaObject, pgSpec)
 			if err != nil {
 				log.Warnf("Sync PodGroup %v: %v", jobKey, err)
+				return nil
+			}
+
+			if pg == nil || pg.Status.Phase == "" || pg.Status.Phase == v1beta1.PodGroupPending {
```
terrytangyuan, line 316: Could you add some code comments on this, similar to issue description in https://github.com/kubeflow/common/issues/129?

shinytang6: done, PTAL

pkg/controller.v1/common/job.go
```diff
@@ -307,9 +307,24 @@ func (jc *JobController) ReconcileJobs(
...
+				log.Warnf("PodGroup %v unschedulable", jobKey)
+				syncReplicas = false
+			}
+
+			if !syncReplicas {
+				now := metav1.Now()
+				jobStatus.LastReconcileTime = &now
+
+				return jc.Controller.UpdateJobStatusInApiServer(job, &jobStatus)
```
gaocegege, line 327: I think we could add some comments here to let us know that we do this update to trigger another reconcile. And maybe in the future the operator should watch PodGroup directly.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #137, opened 2021-06-26
optimize the ttl of job

HeGaoYuan:
my implementation refer to 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/ttlafterfinished/ttlafterfinished_controller.go#L290

### DIFF
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -401,21 +401,29 @@ func (jc *JobController) CleanupJob(runPolicy *apiv1.RunPolicy, jobStatus apiv1.
 		return nil
 	}
 	duration := time.Second * time.Duration(*ttl)
-	if currentTime.After(jobStatus.CompletionTime.Add(duration)) {
+	// todo: Is the jobStatus.CompletionTime maybe nil ?
+	finishTime := jobStatus.CompletionTime
+	expireTime := finishTime.Add(duration)
+	if currentTime.After(expireTime) {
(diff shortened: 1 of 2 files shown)


gaocegege (commented):
/assign @Jeffwan @johnugeorge

gaocegege:

pkg/controller.v1/common/job.go
```diff
@@ -45,32 +45,6 @@ func (jc *JobController) DeletePodsAndServices(runPolicy *apiv1.RunPolicy, job i
 	return nil
 }
 
-func (jc *JobController) cleanupJobIfTTL(runPolicy *apiv1.RunPolicy, jobStatus apiv1.JobStatus, job interface{}) error {
```
johnugeorge, line 48: Which code is referencing this function?

HeGaoYuan: @johnugeorge I don't find any code is referencing the cleanupJobIfTTL function. And I found that the cleanupJobIfTTL function is same with the CleanupJob function( before I edit), so I removed it.

gaocegege: LGTM, it seems that the func is not used now.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

HeGaoYuan:
@terrytangyuan @gaocegege 
But I left a todo:  Is the jobStatus.CompletionTime maybe nil ?

PR #141, opened 2021-07-10
feat(common) add reconciler.v1

zw0610:
#140 

The common repository offers a `controller.v1` package for developers to make kubeflow operators. The controller mode exposes low-level APIs for operators needs high customization requests, which are more about the controller mechanism, such like [pre-processing/post-processing for enqueue/dequeue actions](https://github.com/kubeflow/tf-operator/pull/1315#discussion_r680465216).

However, for more entry-level developers or users who may not that familiar with controller, exposing to many low-level APIs could be confusing. Meanwhile, with effort from controller-runtime, kubebuilder and operator-sdk, the `reconciler` mode is offering a high-level, controller-mechanism-decoupled APIs which are only related to how to meet the resources expectation defined in the declarative API like TFJob, MXNetJob, etc.

To add the `reconciler.v1` package as long as to keep code re-use between the `controller.v1` and `reconciler.v1` packages, `core` package is extracted to be shared between `reconciler.v1` and `controller.v1`.

The `reconciler.v1` defines its API in [reconciler.v1/common/interface.go](https://github.com/kubeflow/common/blob/1d45260761eb3fb5f7ea6ec0a672408f896edc91/pkg/reconciler.v1/common/interface.go) which methods only related to reconciliation exposed.

A base/parent implementation named `KubeflowReconciler` is defined so developers only need to override methods in [reconciler.v1/common/must_customize.go](https://github.com/kubeflow/common/blob/1d45260761eb3fb5f7ea6ec0a672408f896edc91/pkg/reconciler.v1/common/must_customize.go).

- [x] Add `reconciler.v1` package (`controller.v1` is also modified to extract shared functionality)
- [x] Add API docs in terms of comment
- [x] Add unit-test cases for `reconciler.v1`

### DIFF
--- a/pkg/reconciler.v1/common/interface.go
+++ b/pkg/reconciler.v1/common/interface.go
@@ -1,0 +1,260 @@
+// Copyright 2021 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+
+	commonv1 "github.com/kubeflow/common/pkg/apis/common/v1"
+
+	"github.com/go-logr/logr"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ReconcilerUtilInterface defines the abstract interface of reconciler on utility features, such like get event
+// recorder or logger
+type ReconcilerUtilInterface interface {
+	// GetReconcilerName SHOULD be overridden if a new Reconciler is defined. The default implementation returns
+	// "Kubeflow Reconciler"
+	GetReconcilerName() string
+
+	// GetRecorder CAN be overridden to customize EventRecorder
+	GetRecorder() record.EventRecorder
+
+	// GetLogger CAN be overridden to customize logger
+	GetLogger(job client.Object) logr.Logger
+
+	// GetScheme CAN be overridden to customize runtime scheme
+	GetScheme() *runtime.Scheme
+}
+
+// GangSchedulingInterface defines the abstract interface for gang-scheduling related actions, such like get, create or
+// delete PodGroup
+type GangSchedulingInterface interface {
+	// OverrideForGangSchedulingInterface MUST NOT be overridden as it reset ReconcilerUtilInterface
+	OverrideForGangSchedulingInterface(ui ReconcilerUtilInterface)
+
+	// GangSchedulingEnabled CAN be overridden if definition of gang-scheduling enabling changes.
+	GangSchedulingEnabled() bool
+
+	// GetGangSchedulerName CAN be overridden to customize the name of gang scheduler. This name will be used to check
+	// the value of podTemplateSpec.Spec.SchedulerName. For volcano, it is "volcano".
+	GetGangSchedulerName() string
+
+	// GetPodGroupName CAN be overridden to customize the name of PodGroup generated for the job. For example:
+	// podGroupName := fmt.Sprintf("%s-podgroup", job.GetName()) or podGroupName := job.GetName()
+	GetPodGroupName(job client.Object) string
+
+	// GetPodGroupForJob SHOULD be overridden if Group, APIVersion or Kind changes for PodGroup. The PodGroup is
+	// defined in different gang-scheduler as:
+	// Kube-Batch:          "scheduling.incubator.k8s.io/v1alpha1/PodGroup", "scheduling.sigs.dev/v1alpha2/PodGroup"
+	// Volcano:             "scheduling.volcano.sh/v1beta1/PodGroup"
+	// Scheduler-Framework: "scheduling.sigs.k8s.io/v1alpha1/PodGroup"
+	GetPodGroupForJob(ctx context.Context, job client.Object) (client.Object, error)
+
+	// DeletePodGroup SHOULD be overridden if Group, APIVersion and Kind changes for PodGroup.
+	DeletePodGroup(ctx context.Context, job client.Object) error
+
+	// ReconcilePodGroup CAN be overridden if the logic to reconcile PodGroup changes.
+	ReconcilePodGroup(ctx context.Context, job client.Object, runPolicy *commonv1.RunPolicy,
+		replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec) error
+
+	// DecoratePodForGangScheduling SHOULD be overridden if gang scheduler demands Pods associated with PodGroup to be
+	// decorated with specific requests.
+	DecoratePodForGangScheduling(rtype commonv1.ReplicaType, podTemplate *corev1.PodTemplateSpec, job client.Object)
+}
+
+// PodInterface defines the abstract interface for Pod related actions, such like get, create or delete Pod
+type PodInterface interface {
+	// OverrideForPodInterface MUST NOT be overridden as it reset ReconcilerUtilInterface, GangSchedulingInterface, JobInterface
+	OverrideForPodInterface(ui ReconcilerUtilInterface, gi GangSchedulingInterface, ji JobInterface)
+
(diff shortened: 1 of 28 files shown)


Jeffwan:
Let's try to minimize the code changes. If  `pod.go` `services.go` are reusable, then let's try not to replica the changes

zw0610:
> Let's try to minimize the code changes. If `pod.go` `services.go` are reusable, then let's try not to replica the changes

sure. I shall finish the gang part and try to maximize code re-use in the next commit.

zw0610:
@Jeffwan While I'm still working on adding comment and unit-tests for `reconciler.v1` package, you may take a look at this pull request and let me know any concern to the changes especially to the existing `controller.v1` package.

Let me know if I need to rebase it first.

pkg/reconciler.v1/common/interface.go
```diff
@@ -0,0 +1,248 @@
...
+	// DeleteService CAN be overridden to customize how to delete the service of {name} in namespace {ns}.
+	DeleteService(ns string, name string, job client.Object) error
+
+	// DecorateService CAN be overridden to customize this service right before being created
+	DecorateService(rtype commonv1.ReplicaType, podTemplate *corev1.PodTemplateSpec, job client.Object)
+}
+
+// JobInterface defines the abstract interface for Pod related actions, such like get, create or delete TFJob,
+// PyTorchJob or KFJob
+type JobInterface interface {
```
Jeffwan, line 155: Seems it brings more interfaces. I notice this change is very large and includes lots of refactoring. Maybe it's better to add more details in the issue for training leads to review? 

It would be great to cover following topics
1. What's the advantage of using `reconciler.v1` over `controller.v1`? 
2. What's the user facing changes?  for example, if user wants to build an operator, how this solution benefit their development? 

zw0610: For questions here, I updated https://github.com/kubeflow/common/pull/141#issue-687173348 to elaborate the benefit of `reconciler.v1`.

But from my perspective, I would keep both (reconciler and controller) for developers/users at different level.

johnugeorge:
In general, do we need to low level api code  in longer run?  Can we discuss in next meeting before merge?

Jeffwan:
/hold 

@johnugeorge Sure, this is a big change requires user (operator) facing changes. Let's have an agreement in the meeting first.

zw0610:
> In general, do we need to low level api code in longer run? Can we discuss in next meeting before merge?

I think the `Reconciler` struct exposes limited APIs that are sufficient for most operators development. While there might always be some customization supported only in `Controller` mode, especially for users with private deploy, these two mode should be 

The concern on keeping both `reconciler.v1` and `controller.v1`, I believe, is about the effort to maintain features and bug-fix for both package. I would suggest the following steps:

1. add `reconciler.v1` package to kubeflow/common repository
2. convert controllers in kubeflow/tf-operator based on kubeflow/common `reconciler.v1`
3. deprecate `controller.v1`

Because the scheduling for the community meeting next week (August 11/12th) is not so friendly to my local time, I can attend if the discussion is quite critical. Otherwise, could we discuss on August 25th?

go.mod
```diff
@@ -23,4 +26,6 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.19.9
 	k8s.io/client-go => k8s.io/client-go v0.19.9
 	k8s.io/code-generator => k8s.io/code-generator v0.19.9
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.7.2
+	github.com/prometheus/client_golang v1.11.0 => github.com/prometheus/client_golang v1.7.1
```
andreyvelich, line 30: Do we need this [replace](https://github.com/kubeflow/common/blob/a19dc4d55450431b1a91dbdcf793e7c58018bf77/go.mod#L24-L30) or we can remove it since we are using `v0.21.3` version of K8s packages ?

zw0610: I'm trying to keep go packages in common and tf-operator consistent. We can discuss this dependency issue on the next community meeting (August 11th).

zw0610: If we are using `v0.21.3`, shall I get rid of other replace items like `k8s.io/client-go => k8s.io/client-go v0.19.9`?

alculquicondor:
This is a long PR. Would it be worth splitting the go.mod upgrades in its own commit or PR?

Any specific parts were you would like my input, given that I'm still unfamiliar with this project.

pkg/reconciler.v1/common/job.go
```diff
@@ -0,0 +1,478 @@
...
+	client.Client
+	ReconcilerUtilInterface
+	PodInterface
+	ServiceInterface
+	GangSchedulingInterface
+	counter *commonutil.Counter
+}
+
+// BareKubeflowJobReconciler returns the pointer of a KubeflowJobReconciler with minimal implementation
+func BareKubeflowJobReconciler(client client.Client) *KubeflowJobReconciler {
```
Jeffwan, line 76: Should we use term like "Base"

pkg/reconciler.v1/common/job_must_customize.go
```diff
@@ -0,0 +1,49 @@
...
+
+	commonv1 "github.com/kubeflow/common/pkg/apis/common/v1"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetJob returns the job that matches the request
+func (r *KubeflowJobReconciler) GetJob(ctx context.Context, req ctrl.Request) (client.Object, error) {
+	panic("implement KubeflowReconciler.GetJob!")
```
Jeffwan, line 28: Does it need to implement all methods? 

These methods will be overridden by upstream reconciler? is there any chance to call these and get panic?

pkg/reconciler.v1/common/job_must_customize.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 The Kubeflow Authors
```
Jeffwan, line 1: Trying to understand name "job_must_customize.go" Is there a better name for some interfaces needed to be implemented?

Jeffwan:
The scope and codes overall looks good to me. Other leads, please leave comments.

go.mod
```diff
@@ -3,24 +3,16 @@ module github.com/kubeflow/common
...
-	k8s.io/api v0.19.9
-	k8s.io/apimachinery v0.19.9
-	k8s.io/client-go v0.19.9
-	k8s.io/code-generator v0.19.9
-	k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
+	github.com/prometheus/client_golang v1.11.0
+	github.com/sirupsen/logrus v1.7.0
+	github.com/stretchr/testify v1.7.0
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
```
Jeffwan, line 12: One more request, could we split the dependency upgrade PR to a separate one?

zw0610: sure.

pkg/reconciler.v1/common/job.go
zw0610: While the `Bare` means only interface independently defined in `JobInterface` is implemented in the returned `KubeflowJobReconciler` pointer, I think the `Base` conveys the same message.

pkg/reconciler.v1/common/job_must_customize.go
zw0610: At this stage, it is designed to call the panic if developers uses `KubeflowJobReconciler` directly. It forces developers to override these methods.

However, after the generic `KubeflowJob` is introduced in tf-operator or we call unified training operator, these `panic` functions will be changed into warning or logging info.

pkg/reconciler.v1/common/job_must_customize.go
zw0610: There should be a better name. Maybe we can call it `override.go`. I shall put a readme file right within the `reconciler` package.

pkg/reconciler.v1/common/job_must_customize.go
Jeffwan: If you forces developer to override these methods. Why just not to implement these methods so end user will get compilation error. This way gives user runtime error. is that correct?

zw0610: I see, exactly! Now all these `panic` methods are removed. Meanwhile, the `job_must_customize.go` file is replaced by a `README.md` to inform developers these function should be implemented. A simple example can be found at `test_job/reconciler.v1`

Jeffwan: Thanks for the update. Great!

Jeffwan:
Overall looks good to me. I think this is safe to merge and it doesn't introduce user facing low level JobController. (just some reorganization to extract common codes). 

If we plan to support a CustomJob which can support arbitrary roles. reconciler.v1 would be a good start. For existing frameworks, we can make some plans later. This is not a blocking issue to 1.4 release and beta release of all-in-one operator.

We can probably introduce this to v2 apis & controllers. 

@kubeflow/wg-training-leads  Please have a double check. Let's hold it for two days.

johnugeorge:
I agree with @Jeffwan  For existing frameworks this is not necessary at this point. We can plan for it at a later point

PR #146, opened 2021-07-23
fix: Add uppercase support for GenGeneralName

gaocegege:
In GenGeneralName, we should use lower case.

In the past we do the conversion outside the function since the replicaType is a string. Now its type is replicaType, thus we should do it in the function.

/cc @terrytangyuan @MartinForReal 

Signed-off-by: cegao <cegao@tencent.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+vendor/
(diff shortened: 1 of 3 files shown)


terrytangyuan (approved):
Thanks!

/lgtm
Thanks!

/lgtm
/approve

Issue #149, opened 2021-08-09
Fully-qualified labels migration tasks


alculquicondor:
These tasks must be completed to finalize the migration to fully qualified labels #148 

- [x] Migrate `SyncPdb` to non-deprecated label
- [x] Remove deprecated selector from `FilterPodsForReplicaType`
- [x] Remove deprecated selector from `FilterServicesForReplicaType`
- [x] Remove deprecated labels from API
- [x] Remove deprecated labels usage from pkg/util/labels
- [x] Remove deprecated labels from `GenLabels`

PR #150, opened 2021-08-09
Add fully-qualified labels

### LINKED ISSUE #148: Use fully-qualified labels
Pods created with kubeflow controllers get some labels https://github.com/kubeflow/common/blob/master/pkg/apis/common/v1/constants.go

It is common practice in k8s to add a domain to the label name. This should reduce the chances of collision with user-defined labels. Then, the labels should go from:
- `replica-index` to `kubeflow.org/replica-index`
- `replica-type` to `kubeflow.org/replica-type`
- `replica-group` to `kubeflow.org/replica-group` (although we could take the chance to rename this to `kubeflow.org/operator-name` with possible values `tf-operator`, `mpi-operator` and so on)
- `job-name` to `kubeflow.org/job-name`
- `job-role` to `kubeflow.org/job-role` (how does this differ from `replica-type`?)

This should enhance visibility and make it easier for cluster administrators to track the usage of operators in the cluster.

The way we can implement this is to start adding both existent and new labels to newly created pods, and change the have something like this in the `constants.go` file:

```golang
const ReplicaIndexLabelDeprecated = "replica-index"
const ReplicaIndexLabel           = "kubeflow.org/replica-index"
```

Then, we would remove the Deprecated variables after a few releases

alculquicondor:
Add a label for operator name.
Mark non-fully-qualified labels as Deprecated.

Fixes #148

Pending tasks to be done in future releases #149

### DIFF
--- a/pkg/apis/common/v1/constants.go
+++ b/pkg/apis/common/v1/constants.go
@@ -1,18 +1,40 @@
 package v1
 
 const (
-	// ReplicaIndexLabel represents the label key for the replica-index, e.g. the value is 0, 1, 2.. etc
-	ReplicaIndexLabel = "replica-index"
+	// TODO(#149): Remove deprecated labels.
 
-	// ReplicaTypeLabel represents the label key for the replica-type, e.g. the value is ps , worker etc.
-	ReplicaTypeLabel = "replica-type"
+	// ReplicaIndexLabel represents the label key for the replica-index, e.g. 0, 1, 2.. etc
+	ReplicaIndexLabel = "training.kubeflow.org/replica-index"
 
-	// GroupNameLabel represents the label key for group name, e.g. the value is kubeflow.org
-	GroupNameLabel = "group-name"
+	// ReplicaIndexLabelDeprecated represents the label key for the replica-index, e.g. the value is 0, 1, 2.. etc
+	// DEPRECATED: Use ReplicaIndexLabel
+	ReplicaIndexLabelDeprecated = "replica-index"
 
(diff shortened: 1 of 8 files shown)


gaocegege (commented):
/lgtm

It's clean.

terrytangyuan (approved):
/approve

alculquicondor:
Can we have a new release of this repo?

Jeffwan:
> Can we have a new release of this repo?

Here's the diff. https://github.com/kubeflow/common/compare/v0.3.6...master  I think we can cut one release or wait for more features. Can also use commit in go.mod?

alculquicondor:
I would prefer we have a release than using a commit in go.mod

gaocegege:
@Jeffwan WDYT, should we have a release after the all-in-one operator is merged?

Jeffwan:
@gaocegege @alculquicondor I think it's ok to cut a release. We probably need to define release strategy for this repo. 

I assume changes like #150  are breaking changes? If job controller is released, on going jobs will be affected. 

Do we want to cut 0.3.7 or cut release branch 0.4.0? We didn't cut release branches for this version yet. I think it's time to follow semantic versioning?

alculquicondor:
It's only breaking for the controllers that don't use the common controller. For the ones that do, the new controller still respects the deprecated APIs.
But 0.4.0 makes the most sense

Jeffwan:
Sounds good.  Let's cut `release-0.4` branch and cut 0.4.0 tag for the new release.

PR #151, opened 2021-08-12
fix replicaType key for filtering pods and services

zw0610:
In #150, when updating the selector labels for matching services and pods, key `ReplicaIndexLabel` is introduced while I believe it needs `ReplicaTypeLabel`.


gaocegege (commented):
/assign @alculquicondor

alculquicondor:
/lgtm

Although I added a version with a unit test #152

PR #155, opened 2021-08-29
Use PriorityClassInterface instead of lister

Jeffwan:
Fix this issue in v0.3 branch

https://github.com/kubeflow/tf-operator/issues/1382

The major problem is we didn't initial `PriorityClassLister` after common migration. When it tries to create podGroup and calculate resources, it's panic. 

I've tested with tf-operator master. v1.2.0 should be fixed same way

### DIFF
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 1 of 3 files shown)


Jeffwan:
kubeflow/common master needs this change as well. Since file structure has been changed a little bit. Let's check in the codes later

Jeffwan:
I didn't notice Ce already filed https://github.com/kubeflow/tf-operator/pull/1384. I will close this one.

Jeffwan:
@gaocegege is there a way we can make this consistent between 1.2-branch and master? Seem 1.2 uses lister and master has to use Interface

Jeffwan:
Let me create a new informer from clientset and we can stick to lister here.

gaocegege:
Can you please open an issue t keep track of it?

Issue #159, opened 2021-09-06
Missing methods in reconciler.v1


zw0610:
In https://github.com/kubeflow/common/pull/141, methods implemented as `panic("implement me!")` that are aimed to let developers overriding is removed as the pull request is merged. However, when creating a demo reconciler or even run the test file `pkg/reconciler.v1/common/service_test.go`, it raises error:

```shell
# github.com/kubeflow/common/pkg/reconciler.v1/common
pkg/reconciler.v1/common/reconciler.go:60:24: cannot use jobInter (type *KubeflowJobReconciler) as type JobInterface in assignment:
	*KubeflowJobReconciler does not implement JobInterface (missing ExtractJobStatus method)
pkg/reconciler.v1/common/reconciler.go:66:24: cannot use jobInter (type *KubeflowJobReconciler) as type JobInterface in assignment:
	*KubeflowJobReconciler does not implement JobInterface (missing ExtractJobStatus method)
pkg/reconciler.v1/common/reconciler.go:73:34: cannot use jobInter (type *KubeflowJobReconciler) as type JobInterface in assignment:
	*KubeflowJobReconciler does not implement JobInterface (missing ExtractJobStatus method)
FAIL	command-line-arguments [build failed]
FAIL
```

Simply comment out the `DefaultKubeflowReconciler` method in `pkg/reconciler.v1/common/reconciler.go` does not resolve the problem as new error appears:

```shell
# github.com/kubeflow/common/test_job/reconciler.v1/test_job
test_job/reconciler.v1/test_job/test_job_reconciler.go:54:24: cannot use jobInter (type *"github.com/kubeflow/common/pkg/reconciler.v1/common".KubeflowJobReconciler) as type "github.com/kubeflow/common/pkg/reconciler.v1/common".JobInterface in assignment:
	*"github.com/kubeflow/common/pkg/reconciler.v1/common".KubeflowJobReconciler does not implement "github.com/kubeflow/common/pkg/reconciler.v1/common".JobInterface (missing ExtractJobStatus method)
test_job/reconciler.v1/test_job/test_job_reconciler.go:60:24: cannot use jobInter (type *"github.com/kubeflow/common/pkg/reconciler.v1/common".KubeflowJobReconciler) as type "github.com/kubeflow/common/pkg/reconciler.v1/common".JobInterface in assignment:
[...]

PR #161, opened 2021-09-19
fix: add missing dependencies and missing methods

Jeffwan:
Address  #161  #159

This is related to @zw0610 's reconciler v1 work. 

I am trying to see if the CI can pass and why we didn't detect the failure earlier.


Jeffwan:
em.. Seems travis has been removed? #141 introduces some build failures and I think that's the reason we didn't catch it in presubmit phase..

gaocegege:
I have no idea about it. Should we use github actions?

/lgtm
/approve

Jeffwan:
@gaocegege Not sure what happened to travis. Agree, I create an issue to track it https://github.com/kubeflow/common/issues/162

Issue #162, opened 2021-09-19
Enable GitHub action to build and test code


Jeffwan:
See https://github.com/kubeflow/common/pull/161

#141 introduces some build failures but we don't notice that before we merge the code. Travis somehow was deleted at that time. @gaocegege suggests to use Github actions. 

Let's migrate scripts https://github.com/kubeflow/common/blob/master/.travis.yml to a Github Action manner.

We do need this to guarantee code quality 

/help
/good-first-issue

Jeffwan:
https://github.com/kubeflow/common/runs/3648012266 Interesting, in the latest workflow, I didn't see failure logs..

SumLare:
Well https://status.coveralls.io/ still doesn't look good

Jeffwan:
Yeah, that's the problem

```
bad response status from coveralls: 503
<h2>This website is under heavy load (queue full)</h2><p>We're sorry, too many people are accessing this website at the same time. We're working on this problem. Please try again later.</p>

```

gaocegege:
coveralls crashes sometimes.

Jeffwan:
Let's do this to overcome the issue. Since goveralls is nice to have, we can ignore the failures in CI. I can not believe recovery takes that long.. 

`goveralls || echo "push to coveralls failed"`

Jeffwan:
This can be closed. thanks @SumLare !

PR #164, opened 2021-09-19
Move ci from travis to github actions

SumLare:
#162

Looks like Coveralls api is under maintenance right now, hopefully ci turns green after they finish


SumLare:
@googlebot I signed it!

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
...
+        # Install golangci-lint
+        curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
+
+    - name: Test
+      run: |
+        golangci-lint run --config=linter_config.yaml ./...
+        # Here we run all tests in pkg and we have to use `-ignore`
+        # since goveralls uses `filepath.Match` to match ignore files
+        # and it does not support patterns like `**`.
+        goveralls -service=github -v -package ./... -ignore "test_job/client/*/*.go,test_job/client/*/*/*.go,test_job/client/*/*/*/*.go,test_job/client/*/*/*/*/*.go,test_job/client/*/*/*/*/*/*.go,test_job/client/*/*/*/*/*/*/*.go,test_job/testutil/*.go,test_job/*/*/*/zz_generated.*.go,test_job/*/*/*/*_generated.go,pkg/apis/common/*/zz_generated.*.go,pkg/apis/common/*/*_generated.go"
```
Jeffwan, line 50: Using `goveralls` looks good

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
+on: push
```
Jeffwan, line 1: Let's also target `pull_request` and kick off presubmit job?

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
...
+      GOPATH: ${{ github.workspace }}
+      GO111MODULE: on
+    defaults:
+      run:
+        working-directory: ${{ env.GOPATH }}/src/github.com/kubeflow/common
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+        path: ${{ env.GOPATH }}/src/github.com/kubeflow/common
```
Jeffwan, line 16: I can not remember the default behavior here. Without specifying the ref, does it by default 
1. fetch HEAD for `pull_request`? 
2. fetch master for `push`

Jeffwan:
@SumLare Thanks for taking this story. Can you help look at the comments?

.github/workflows/build.yml
SumLare: > The branch, tag or SHA to checkout. When checking out the repository that
> triggered a workflow, this defaults to the reference or SHA for that event.
> Otherwise, uses the default branch.

SumLare: So I guess it fetches HEAD of current branch in both of those events, at least it's also how I remember. I can double check though

SumLare: Ok yeah, as per docs:

1. [pull_request](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) -> Last merge commit on the GITHUB_REF branch (master)
2. [push](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push) -> Commit pushed

Jeffwan:
Thanks for the contribution.

/lgtm
/approve

Jeffwan (approved):

PR #165, opened 2021-09-21
Make goveralls step in ci optional

SumLare:
#162

Looks like they are back actually


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #163, opened 2021-09-19
WIP: Revert #135: change rtype to commonv1.ReplicaType

Jeffwan:
I filed this PR https://github.com/kubeflow/common/pull/158 against release-0.3 branch but didn't check in code against master at that time.  Please check details in #158 for the reason why I decide to revert this change.  

> A few things to note

1. Since CI was removed for unknown reason. I manually verify the code quality

```
➜  common git:(revert_135_new) ✗ go build ./...
➜  common git:(revert_135_new) ✗ go test ./...
?   	github.com/kubeflow/common/pkg/apis/common/v1	[no test files]
ok  	github.com/kubeflow/common/pkg/controller.v1/common	0.018s
ok  	github.com/kubeflow/common/pkg/controller.v1/control	0.028s
ok  	github.com/kubeflow/common/pkg/controller.v1/expectation	0.015s
?   	github.com/kubeflow/common/pkg/core	[no test files]
ok  	github.com/kubeflow/common/pkg/reconciler.v1/common	0.029s
ok  	github.com/kubeflow/common/pkg/util	0.018s
?   	github.com/kubeflow/common/pkg/util/k8sutil	[no test files]
ok  	github.com/kubeflow/common/pkg/util/labels	0.021s
?   	github.com/kubeflow/common/pkg/util/signals	[no test files]
ok  	github.com/kubeflow/common/pkg/util/train	0.014s
?   	github.com/kubeflow/common/test_job/apis/test_job/v1	[no test files]
?   	github.com/kubeflow/common/test_job/client/clientset/versioned	[no test files]
?   	github.com/kubeflow/common/test_job/client/clientset/versioned/fake	[no test files]
?   	github.com/kubeflow/common/test_job/client/clientset/versioned/scheme	[no test files]
?   	github.com/kubeflow/common/test_job/client/clientset/versioned/typed/test_job/v1	[no test files]
?   	github.com/kubeflow/common/test_job/client/clientset/versioned/typed/test_job/v1/fake	[no test files]
?   	github.com/kubeflow/common/test_job/client/informers/externalversions	[no test files]
?   	github.com/kubeflow/common/test_job/client/informers/externalversions/internalinterfaces	[no test files]
?   	github.com/kubeflow/common/test_job/client/informers/externalversions/test_job	[no test files]
?   	github.com/kubeflow/common/test_job/client/informers/externalversions/test_job/v1	[no test files]
?   	github.com/kubeflow/common/test_job/client/listers/test_job/v1	[no test files]
[...]

PR #166, opened 2021-09-28
simplify reconciler

zw0610:
To support `GenericJob` in tf-operator, a kind of Job independent from any deep learning frameworks, this pull request introduces the following modification to `reconciler.v1` package:

1. Simplify multiple reconciler names (by dropping redundant words like `Kubeflow`)
2. Fix `PodList`, `ServiceList` converting issue
3. Fix default `GetReconcilerName` returned value
4. Simplify `reconciler.v1` package by moving the `KubeflowReconciler` to tf-operator https://github.com/kubeflow/tf-operator/pull/1398
3. Simplify `TestReconciler` construction function


pkg/reconciler.v1/common/service.go
```diff
@@ -42,23 +42,23 @@ var (
 	})
 )
 
-// KubeflowServiceReconciler defines a Service Reconciler for KubeflowJob
-type KubeflowServiceReconciler struct {
+// ServiceReconciler defines a Service Reconciler for KubeflowJob
```
terrytangyuan, line 45: Still mentions KubeflowJob here

pkg/reconciler.v1/common/README.md
```diff
@@ -6,19 +6,19 @@ To use the reconciler, following methods must be overridden according to the API
 
 ```go
 // GetJob returns the job that matches the request
-func (r *KubeflowJobReconciler) GetJob(ctx context.Context, req ctrl.Request) (client.Object, error)
+func (r *JobReconciler) GetJob(ctx context.Context, req ctrl.Request) (client.Object, error)
```
terrytangyuan, line 9: What about something like Training or TrainingJob since we are always dealing with model trainings here. Job seems very generic.

pkg/reconciler.v1/common/service.go
zw0610: Thanks for pointing out. I changed `KubeflowJob` to `generic training job`.

pkg/reconciler.v1/common/README.md
zw0610: `TrainingJob` sounds good to me. As the common library is designed for all kinds of controllers, I would suggest renaming the `GenericJob` to `TrainingJob` in the pull request I mentioned above for tf-operator (https://github.com/kubeflow/tf-operator/pull/1398). What do you think?

terrytangyuan: SGTM

Jeffwan:
Sorry for the delay. Let me merge #163 first before accepting any new changes. This is really annoying.

zw0610:
@Jeffwan could we move this pr forward as the revert string issue is resolved now.

terrytangyuan (approved):
LGTM
/assign @Jeffwan

pkg/reconciler.v1/common/utils.go
```diff
@@ -22,9 +22,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const ReconcilerName = "Kubeflow Reconciler"
+const ReconcilerName = "common-reconciler"
 
-// GetReconcilerName returns the name of this reconciler, which is "Kubeflow Reconciler"
+// GetReconcilerName returns the name of this reconciler, which is "common-reconciler"
 func (r *ReconcilerUtil) GetReconcilerName() string {
```
Jeffwan, line 28: "Kubeflow Reconciler" won't work because it's not a word?

pkg/reconciler.v1/common/interface.go
```diff
@@ -239,21 +239,3 @@ type JobInterface interface {
 	// PastActiveDeadline CAN be overridden to customize how to determine if this job has past activate deadline.
 	PastActiveDeadline(runPolicy *commonv1.RunPolicy, jobStatus *commonv1.JobStatus) bool
 }
-
-// KubeflowReconcilerInterface defines the abstract interface for a base reconciler for kubeflow jobs.
-type KubeflowReconcilerInterface interface {
```
Jeffwan, line 244: What's the motivation to move `KubeflowReconcilerInterface` to `training-operator`?

pkg/reconciler.v1/common/service.go
```diff
@@ -72,40 +72,40 @@ func (r *KubeflowServiceReconciler) OverrideForServiceInterface(ui ReconcilerUti
...
-func (r *KubeflowServiceReconciler) GetServicesForJob(ctx context.Context, job client.Object) ([]*corev1.Service, error) {
+func (r *ServiceReconciler) GetServicesForJob(ctx context.Context, job client.Object) ([]*corev1.Service, error) {
 	svcList := &corev1.ServiceList{}
 	err := r.List(ctx, svcList, client.MatchingLabels(r.GenLabels(job.GetName())))
 	if err != nil {
 		return nil, err
 	}
 
 	var svcs []*corev1.Service = nil
-	for _, svc := range svcList.Items {
```
Jeffwan, line 89: em. what's the problem of original way?

pkg/reconciler.v1/common/service.go
```diff
@@ -72,40 +72,40 @@ func (r *KubeflowServiceReconciler) OverrideForServiceInterface(ui ReconcilerUti
...
 // GetServicesForJob returns all services associated with this job
-func (r *KubeflowServiceReconciler) GetServicesForJob(ctx context.Context, job client.Object) ([]*corev1.Service, error) {
+func (r *ServiceReconciler) GetServicesForJob(ctx context.Context, job client.Object) ([]*corev1.Service, error) {
 	svcList := &corev1.ServiceList{}
 	err := r.List(ctx, svcList, client.MatchingLabels(r.GenLabels(job.GetName())))
 	if err != nil {
 		return nil, err
 	}
 
 	var svcs []*corev1.Service = nil
```
Jeffwan, line 88: Can we do `var svcs []*corev1.Service`?  it's a little bit weird to explicitly assign nil here.

Jeffwan:
@zw0610 sorry for late response. Please have a check on above comments

pkg/reconciler.v1/common/utils.go
zw0610: Yes. When generating the default labels, the returned string from `GetReconcilerName` will be used as the value for key `commonv1.OperatorNameLable` (`""training.kubeflow.org/operator-name""`).

pkg/reconciler.v1/common/interface.go
zw0610: When writing the test code, `test_reconciler`, which could be considered as one example of how the `common.reconciler` will be used, I found that:

1. Such interface will not be used for any abstraction when implementing a real reconciler
2. Embedding modularized interfaces in `KubeflowReconcilerInterface` conflicts embedding them in the real reconciler struct.

pkg/reconciler.v1/common/service.go
zw0610: sure. also changed for `var pods []*corev1.Pod`.

pkg/reconciler.v1/common/service.go
zw0610: It seems the old way will not give a new address for `svc` so when we take the pointer of `svc`, it remains the same. Please see the screenshot below.

Screen Shot 2021-12-12 at 20 14 53

pkg/reconciler.v1/common/interface.go
Jeffwan: I think it make sense to me. current testing integration between common and training-operator is a headache.

pkg/reconciler.v1/common/service.go
Jeffwan: I see. yeah.  This pointer points to the shared memory location that each of the slice items are being copied into.
All the pointers created using `&svc` point to the same memory location.

Current way looks good to me.

Jeffwan:
/lgtm
/hold

This change looks good to me. Let's see if anyone else have further feedbacks. /cc @kubeflow/wg-training-leads

Jeffwan (approved):

zw0610:
Any feedback before we can merge this pr? @kubeflow/wg-training-leads

PR #168, opened 2021-10-09
Enforce backOffLimit in ExitCode restartPolicy

### LINKED ISSUE #167: Limit the number of restarts under ExitCode restartPolicy
In the current [implementation](https://github.com/kubeflow/common/blob/master/pkg/core/job.go#L86-L89), the `runPolicy.BackoffLimit` is only applicable on `OnFailure` and `Always` restart policies. I was wondering if there's a reason it's not supported on the `ExitCode` restart policy. Often in the case of OOM, the job exits with an exit code of 137 which is rightfully a retriable error. However, it will keep restarting indefinitely since the policy is not covered by the `BackOffLimit`. The behavior could be to honor the backOffLimit if present, else it keeps retrying indefinitely?

I am happy to submit a PR if you think this is a reasonable change.

goyalankit:
The `runPolicy.backOffLimit` allows one to specify the sum total of number of times pods are allowed to restart before giving up. This change allows the `backOffLimit` to be applied when the `restartPolicy` is set to `ExitCode`

closes #167

### DIFF
--- a/pkg/core/job.go
+++ b/pkg/core/job.go
@@ -74,7 +74,7 @@ func PastActiveDeadline(runPolicy *apiv1.RunPolicy, jobStatus apiv1.JobStatus) b
 }
 
 // PastBackoffLimit checks if container restartCounts sum exceeds BackoffLimit
-// this method applies only to pods with restartPolicy == OnFailure or Always
+// this method applies only to pods when restartPolicy is one of OnFailure, Always or ExitCode
 func PastBackoffLimit(jobName string, runPolicy *apiv1.RunPolicy,
 	replicas map[apiv1.ReplicaType]*apiv1.ReplicaSpec, pods []*v1.Pod,
 	podFilterFunc func(pods []*v1.Pod, replicaType apiv1.ReplicaType) ([]*v1.Pod, error)) (bool, error) {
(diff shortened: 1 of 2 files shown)


Jeffwan:
This is straightforward. The changes look good to me.  /cc @gaocegege for confirmation. 

@goyalankit  Once this is merged. can you cherry-pick the change to release-0.3 branch as well?

pkg/core/job.go
```diff
@@ -83,8 +83,8 @@ func PastBackoffLimit(jobName string, runPolicy *apiv1.RunPolicy,
 	}
 	result := int32(0)
 	for rtype, spec := range replicas {
-		if spec.RestartPolicy != apiv1.RestartPolicyOnFailure && spec.RestartPolicy != apiv1.RestartPolicyAlways {
-			log.Warnf("The restart policy of replica %v of the job %v is not OnFailure or Always. Not counted in backoff limit.", rtype, jobName)
+		if spec.RestartPolicy != apiv1.RestartPolicyOnFailure && spec.RestartPolicy != apiv1.RestartPolicyAlways && spec.RestartPolicy != apiv1.RestartPolicyExitCode {
+			log.Warnf("The restart policy of replica %v of the job %v is not OnFailure, Always or ExitCode. Not counted in backoff limit.", rtype, jobName)
```
gaocegege, line 87: nit: Should we use %v or %s? I am not sure about it.

gaocegege (approved):
/lgtm

PR #172, opened 2021-11-06
Fix openapi generate rule violation

pugangxa:
As mentioned in https://github.com/kubeflow/mpi-operator/pull/434#issuecomment-962420580

```
openapi-gen --input-dirs github.com/kubeflow/mpi-operator/v2/pkg/apis/kubeflow/v2beta1,github.com/kubeflow/common/pkg/apis/common/v1 --output-package github.com/kubeflow/mpi-operator/v2/pkg/apis/kubeflow/v2beta1 --go-header-file hack/boilerplate/boilerplate.go.txt
API rule violation: list_type_missing,github.com/kubeflow/common/pkg/apis/common/v1,JobStatus,Conditions
```

This PR fix this issue.

### DIFF
--- a/pkg/apis/common/v1/types.go
+++ b/pkg/apis/common/v1/types.go
@@ -22,6 +22,7 @@ import (
 // +k8s:deepcopy-gen=true
 // JobStatus represents the current observed state of the training Job.
 type JobStatus struct {
+	// +listType=set
 	// Conditions is an array of current observed job conditions.
 	Conditions []JobCondition `json:"conditions"`
 
--- a/pkg/apis/common/v1/openapi_generated.go
+++ b/pkg/apis/common/v1/openapi_generated.go
@@ -364,6 +364,11 @@ func schema_pkg_apis_common_v1_JobStatus(ref common.ReferenceCallback) common.Op
 				Type:        []string{"object"},
 				Properties: map[string]spec.Schema{
 					"conditions": {
+						VendorExtensible: spec.VendorExtensible{
+							Extensions: spec.Extensions{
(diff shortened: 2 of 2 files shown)


terrytangyuan (approved):
/lgtm
/approve

pkg/apis/common/v1/types.go
```diff
@@ -22,6 +22,7 @@ import (
 // +k8s:deepcopy-gen=true
 // JobStatus represents the current observed state of the training Job.
 type JobStatus struct {
+	// +listType=set
```
alculquicondor, line 25: this should be a map, with listMapKey=type

pugangxa: I'm. afraid there can be conflicts if it's map with `listMapKey=type`. For example, there can be multiple items with the same JobCondition like `Running`.  I think there can't be duplicated items so maybe set is suitable?

alculquicondor: Why would there be multiple conditions for the same type?

pugangxa: Condition is a list and it's open to user to set so it may happen though it's not recommended. But you are right I checked code for JobStatus it's impossible this way. I can modify it then.

alculquicondor: The JobStatus is actually not up to the user, it's entirely controlled by the operators. And if our operators put the same condition more than once, it would be a bug. But generally, lists could act as maps if we decide to.

pugangxa: OK, send another PR https://github.com/kubeflow/common/pull/174 to address this.

PR #178, opened 2021-11-26
feat(job): Add a check to avoid null pointer exception

gaocegege:

/cc @qiankunli

Signed-off-by: Ce Gao <ce.gao@outlook.com>


terrytangyuan (approved):
/lgtm

PR #181, opened 2021-12-09
feat: Add successpolicy

gaocegege:
SuccessPolicy is used in both PyTorchJob and TFJob. Thus I propose to add it in common.

Signed-off-by: cegao <ce.gao@outlook.com>

### DIFF
--- a/pkg/apis/common/v1/openapi_generated.go
+++ b/pkg/apis/common/v1/openapi_generated.go
@@ -509,6 +509,12 @@ func schema_pkg_apis_common_v1_RunPolicy(ref common.ReferenceCallback) common.Op
 							Format:      "",
 						},
 					},
+					"successPolicy": {
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"string"},
+							Format: "",
+						},
+					},
 					"ttlSecondsAfterFinished": {
 						SchemaProps: spec.SchemaProps{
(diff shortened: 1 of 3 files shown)


terrytangyuan (commented):
This is great. Although I wonder if we can use this opportunity to re-think the approach. Instead of enums, would a label selector similar to what Argo Workflows is doing here be more flexible and allows more fine-grained control over the success conditions/policies? https://github.com/argoproj/argo-workflows/blob/master/examples/k8s-kubeflow-jobs.yaml#L24-L25

gaocegege:
It is used in Katib. I think it works, but I think we should support successPolicy to keep API consistency.

cc @andreyvelich

andreyvelich:
Yes, we are using [`successCondition`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L227-L230) with GSON format in our APIs to define condition for Katib Trial's Workers, similar to Argo Workflows.
Probably, we can think how to use it in Training Operators.

terrytangyuan:
Let's use https://github.com/kubeflow/training-operator/issues/1507 to track and discuss separately.

terrytangyuan (approved):
This is LGTM for consistency. Others PTAL.

zw0610:
LGTM. Meanwhile, could you add descriptions for `SchedulingPolicy` and `SuccessPolicyAllWorkers` to explain the expected behavior?

gaocegege:
SGTM

PR #187, opened 2022-04-08

georgkaleido:
If a pod is in phase failure we have to create a new one.
Currently it was assumed the pod would restart due to a RestartPolicy on the pod level
This doesn't work if the pod fails for a system reason.


georgkaleido:
Closing. Reopening with different email for CLA

PR #192, opened 2022-05-12
Add run policy to suspend tasks

PeterChg:
Add the status of the partial success of the task


PeterChg:
> Where is this being used? I don't recall any suspend/resume feature in current controllers.

This is often desired when cluster resources are limited and a higher priority Job needs to execute in the place of another Job.

Arhell (commented):
Please sign CLA

PR #193, opened 2022-05-17
add job suspend run Policy

PeterChg:
add job partial success status


terrytangyuan (commented):
I am not sure if this is a common use case. Could you elaborate?

PeterChg:
> use

The ability to suspend and resume Jobs is often desired when cluster resources are limited and a higher priority Job needs to execute in the place of another Job. 
According to the kubeflow/training-operator project architecture, we need to modify kubeflow/common project first.

PeterChg:
> > use



> I am not sure if this is a common use case. Could you elaborate?



> I am not sure if this is a common use case. Could you elaborate?



> The ability to suspend and resume Jobs is often desired when cluster resources are limited and a higher priority Job needs to execute in the place of another Job.
> According to the kubeflow/training-operator project architecture, we need to modify kubeflow/common project first.



> I am not sure if this is a common use case. Could you elaborate?

The ability to suspend and resume Jobs is often desired when cluster resources are limited and a higher priority Job needs to execute in the place of another Job.
According to the kubeflow/training-operator project architecture, we need to modify kubeflow/common project first.

PeterChg:
> I am not sure if this is a common use case. Could you elaborate?

The ability to suspend and resume Jobs is often desired when cluster resources are limited and a higher priority Job needs to execute in the place of another Job.
According to the kubeflow/training-operator project architecture, we need to modify kubeflow/common project first.

terrytangyuan:
What are the changes you are trying to make to training operator?

PeterChg:
> What are the changes you are trying to make to training operator?

add some logic in pytorch job lifecycle, delete pods when job suspened, create pods when job resumed. optimize pytorchjob status management module, keep the function right after adding suspend/resume state.

terrytangyuan:
> > What are the changes you are trying to make to training operator?
> 
> add some logic in pytorch job lifecycle, delete pods when job suspened, create pods when job resumed. optimize pytorchjob status management module, keep the function right after adding suspend/resume state.

I am not sure if suspend is common in distributed training jobs. There will be side effects depending on the training framework, especially when pods are deleted and recreated.

alculquicondor:
This is not about the training job itself.
This is about a cluster having scarce resources. If there is a higher priority job that needs the resources, `suspend` provides a way to free those resources. The training job will have a chance to checkpoint, if they have the support for it, otherwise just fail and they will be retried later.

PR #190, opened 2022-04-08
Fix exitCode check

georgkaleido:
The exitCode check works differently then described in the [API](https://github.com/kubeflow/common/blob/master/pkg/apis/common/v1/types.go#L163-L168)

In addition the document exitCodes for Docker will not appear in a Kubernetes pod status (anymore?)

PR #195, opened 2022-06-08

yoanisgil:
If a pod is in phase failure we have to create a new one. Currently it was assumed the pod would restart due to a RestartPolicy on the pod level. This doesn't work if the pod fails for a system reason.


johnugeorge:
Can you add a test case? 

I see a related PR for exit code - https://github.com/kubeflow/common/pull/190

/cc @gaocegege @terrytangyuan

yoanisgil:
No need to merge this one as https://github.com/kubeflow/common/pull/189 was already merged.

PR #197, opened 2022-12-17
Changing label selector type for HPA

johnugeorge:
As per docs, https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

`labelSelectorPath` referred by `Scale.Status.Selector` needs to be a string

`The field pointed by this JSON path must be a string field (not a complex selector struct) which contains a serialized label selector in string form.`

Currently, HPA creation fails because of this reason.  Related (https://github.com/kubeflow/training-operator/issues/1645, https://github.com/kubeflow/training-operator/issues/1626)

This change needs changes in training-operator as well which will be fixed in a different PR.

/assign @gaocegege 


gaocegege:
/lgtm

Thanks for the fix!

terrytangyuan (approved):
/lgtm

PR #199, opened 2022-12-20
Install Goveralls using go install

johnugeorge:
Ref: https://go.dev/doc/go-get-install-deprecation

Use go install instead of go get


terrytangyuan (approved):
/lgtm

PR #202, opened 2023-01-11
fix：add Logger for podgroup

tsuiot:
this pull request is 1/2 fix bug of training-operator`s reconcile event loop.

### DIFF
--- a/pkg/util/logger.go
+++ b/pkg/util/logger.go
@@ -16,6 +16,7 @@ package util
 
 import (
 	"strings"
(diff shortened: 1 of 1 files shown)


tenzen-y:
@tsuiot Thanks for reporting this and submitting this patch.
Probably, that bug will be fixed by #1730 since we need to use generic predicates for PodGroup, the same as other controllers.

johnugeorge:
Can we close this PR

PR #209, opened 2023-01-29
Make scheduler-plugins the default gang scheduler

Syulin7:
**What this PR does / why we need it:**

Now supports many gang schedulers(volcano, scheduler-plugins). This PR supports koordinator gang scheduler. 

Related: https://github.com/kubeflow/training-operator/issues/1746

### DIFF
--- a/pkg/controller.v1/common/job.go
+++ b/pkg/controller.v1/common/job.go
@@ -275,7 +275,7 @@ func (jc *JobController) ReconcileJobs(
 					pg = volcanoPodGroup
 					return nil
 				}
-			case GangSchedulerSchedulerPlugins:
+			default:
 				pgSpecFill = func(pg metav1.Object) error {
 					schedulerPluginsPodGroup, match := pg.(*schedulerpluginsv1alpha1.PodGroup)
 					if !match {
--- a/pkg/controller.v1/common/job_controller.go
+++ b/pkg/controller.v1/common/job_controller.go
@@ -69,8 +69,9 @@ var (
 type GangScheduler string
 
 const (
-	GangSchedulerNone             GangScheduler = "None"
-	GangSchedulerVolcano          GangScheduler = "volcano"
+	GangSchedulerNone    GangScheduler = "None"
+	GangSchedulerVolcano GangScheduler = "volcano"
+	// GangSchedulerSchedulerPlugins Using this scheduler name or any scheduler name different than volcano uses the scheduler-plugins PodGroup
 	GangSchedulerSchedulerPlugins GangScheduler = "scheduler-plugins"
 )
 
@@ -81,7 +82,7 @@ type JobControllerConfiguration struct {
 }
 
 func (c *JobControllerConfiguration) EnableGangScheduling() bool {
-	return c.GangScheduling != GangSchedulerNone
+	return c.GangScheduling != "" && c.GangScheduling != GangSchedulerNone
 }
 
 // JobController abstracts other operators to manage the lifecycle of Jobs.
@@ -170,16 +171,16 @@ var GenVolcanoSetupFunc = func(vci volcanoclient.Interface) GangSchedulingSetupF
 	}
 }
 
-var GenSchedulerPluginsSetupFunc = func(c client.Client) GangSchedulingSetupFunc {
+var GenSchedulerPluginsSetupFunc = func(c client.Client, gangSchedulerName string) GangSchedulingSetupFunc {
 	return func(jc *JobController) {
-		jc.Config.GangScheduling = GangSchedulerSchedulerPlugins
-		jc.PodGroupControl = control.NewSchedulerPluginsControl(c)
+		jc.Config.GangScheduling = GangScheduler(gangSchedulerName)
+		jc.PodGroupControl = control.NewSchedulerPluginsControl(c, gangSchedulerName)
 	}
 }
 
 var GenNonGangSchedulerSetupFunc = func() GangSchedulingSetupFunc {
 	return func(jc *JobController) {
-		jc.Config.GangScheduling = GangSchedulerNone
+		jc.Config.GangScheduling = ""
 		jc.PodGroupControl = nil
 	}
 }
@@ -210,7 +211,7 @@ func NewJobController(
 
 	jc := JobController{
 		Controller:     controllerImpl,
-		Config:         JobControllerConfiguration{GangScheduling: GangSchedulerNone},
+		Config:         JobControllerConfiguration{},
 		PodControl:     podControl,
 		ServiceControl: serviceControl,
 		KubeClientSet:  kubeClientSet,
--- a/pkg/controller.v1/common/pod.go
+++ b/pkg/controller.v1/common/pod.go
@@ -421,7 +421,7 @@ func (jc *JobController) createNewPod(job interface{}, rt string, index int, spe
 
 	// if gang-scheduling is enabled:
 	// 1. if user has specified other scheduler, we report a warning without overriding any fields.
-	// 2. if no SchedulerName is set for pods, then we set the SchedulerName to "volcano".
+	// 2. if no SchedulerName is set for pods, we set the SchedulerName to gang-scheduler-name.
 	if jc.Config.EnableGangScheduling() {
 		if isCustomSchedulerSet(replicas, jc.PodGroupControl.GetSchedulerName()) {
 			errMsg := "Another scheduler is specified when gang-scheduling is enabled and it will not be overwritten"
--- a/pkg/controller.v1/control/podgroup_control.go
+++ b/pkg/controller.v1/control/podgroup_control.go
@@ -125,10 +125,15 @@ var _ PodGroupControlInterface = &VolcanoControl{}
 
 // SchedulerPluginsControl is the  implementation of PodGroupControlInterface with scheduler-plugins.
 type SchedulerPluginsControl struct {
-	Client client.Client
+	Client        client.Client
+	SchedulerName string
 }
 
 func (s *SchedulerPluginsControl) DecoratePodTemplateSpec(pts *corev1.PodTemplateSpec, job metav1.Object, _ string) {
+	if len(pts.Spec.SchedulerName) == 0 {
+		pts.Spec.SchedulerName = s.GetSchedulerName()
+	}
+
 	if pts.Labels == nil {
 		pts.Labels = make(map[string]string)
 	}
@@ -136,12 +141,12 @@ func (s *SchedulerPluginsControl) DecoratePodTemplateSpec(pts *corev1.PodTemplat
 }
 
 func (s *SchedulerPluginsControl) GetSchedulerName() string {
-	return ""
+	return s.SchedulerName
 }
 
 // NewSchedulerPluginsControl returns a SchedulerPluginsControl
-func NewSchedulerPluginsControl(c client.Client) PodGroupControlInterface {
-	return &SchedulerPluginsControl{Client: c}
+func NewSchedulerPluginsControl(c client.Client, schedulerName string) PodGroupControlInterface {
+	return &SchedulerPluginsControl{Client: c, SchedulerName: schedulerName}
 }
 
 func (s *SchedulerPluginsControl) DelayPodCreationDueToPodGroup(pg metav1.Object) bool {
--- a/pkg/reconciler.v1/common/gang_scheduler_framework.go
+++ b/pkg/reconciler.v1/common/gang_scheduler_framework.go
@@ -38,6 +38,7 @@ type SchedulerFrameworkReconciler struct {
 	BaseGangReconciler
 	ReconcilerUtilInterface
 	client.Client
+	SchedulerName string
 }
 
 func BareSchedulerFrameworkReconciler(client client.Client, bgReconciler *BaseGangReconciler, enabled bool) *SchedulerFrameworkReconciler {
@@ -57,9 +58,9 @@ func (r *SchedulerFrameworkReconciler) OverrideForGangSchedulingInterface(ui Rec
 	}
 }
 
-// GetGangSchedulerName returns the name of Gang Scheduler will be used, which is "" for SchedulerFrameworkReconciler
+// GetGangSchedulerName returns the name of Gang Scheduler will be used.
 func (r *SchedulerFrameworkReconciler) GetGangSchedulerName() string {
-	return ""
+	return r.SchedulerName
 }
 
 // GangSchedulingEnabled returns if gang-scheduling is enabled for all jobs
@@ -164,6 +165,10 @@ func (r *SchedulerFrameworkReconciler) DecoratePodForGangScheduling(
 	podTemplate *corev1.PodTemplateSpec,
 	job client.Object,
 ) {
+	if podTemplate.Spec.SchedulerName == "" {
+		podTemplate.Spec.SchedulerName = r.GetGangSchedulerName()
+	}
+
 	if podTemplate.Labels == nil {
 		podTemplate.Labels = make(map[string]string)
 	}


johnugeorge:
Thanks @Syulin7 We are in the process of training operator release(currently rc0).  Since kubeflow/common dependency has not branched out, we have to wait a bit to have the final release and then merge this one. 


/cc @tenzen-y  Can you review this?

tenzen-y:
> Thanks @Syulin7 We are in the process of training operator release(currently rc0). Since kubeflow/common dependency has not branched out, we have to wait a bit to have the final release and then merge this one.
> 
> /cc @tenzen-y Can you review this?

Sure. 
First, I will ask questions about the Koordinator at https://github.com/kubeflow/training-operator/issues/1746.

pkg/controller.v1/control/podgroup_control.go
```diff
@@ -193,3 +193,77 @@ func (s *SchedulerPluginsControl) CreatePodGroup(podGroup client.Object) error {
...
+func (s *KoordinatorControl) CreatePodGroup(podGroup client.Object) error {
+	pg := podGroup.(*schedulerpluginsv1alpha1.PodGroup)
+	err := s.Client.Create(context.TODO(), pg, &client.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to create a PodGroup, '%v': %v", klog.KObj(pg), err)
+	}
+	return nil
+}
+
+var _ PodGroupControlInterface = &KoordinatorControl{}
```
tenzen-y, line 269:
```suggestion
// KoordinatorControl is the  implementation of PodGroupControlInterface with scheduler-plugins.
type KoordinatorControl struct {
	*SchedulerPluginsControl
}

var _ PodGroupControlInterface = &KoordinatorControl{}

func (k *KoordinatorControl) DecoratePodTemplateSpec(pts *corev1.PodTemplateSpec, job metav1.Object, _ string) {
	if len(pts.Spec.SchedulerName) == 0 {
		pts.Spec.SchedulerName = k.GetSchedulerName()
	}
	if pts.Labels == nil {
		pts.Labels = make(map[string]string)
	}
	pts.Labels[schedulerpluginsv1alpha1.PodGroupLabel] = job.GetName()
}

func (k *KoordinatorControl) GetSchedulerName() string {
	return "koord-scheduler"
}
```

pkg/reconciler.v1/common/gang_koordinator.go
```diff
@@ -0,0 +1,195 @@
...
+	minMember int32,
+	replicas map[commonv1.ReplicaType]*commonv1.ReplicaSpec,
+) *corev1.ResourceList {
+	return controllerv1.CalcPGMinResources(minMember, replicas,
+		func(pc string) (*schedulingv1.PriorityClass, error) {
+			priorityClass := &schedulingv1.PriorityClass{}
+			err := r.Get(context.TODO(), types.NamespacedName{Name: pc}, priorityClass)
+			return priorityClass, err
+		})
+}
```
tenzen-y, line 195:
```suggestion
type KoordinatorReconciler struct {
	*SchedulerFrameworkReconciler
}

func BareKoordinatorReconciler(client client.Client, bgReconciler *BaseGangReconciler, enabled bool) *KoordinatorReconciler {
	if bgReconciler == nil {
		bgReconciler = &BaseGangReconciler{}
	}
	bgReconciler.Enabled = enabled
	return &KoordinatorReconciler{SchedulerFrameworkReconciler: BareSchedulerFrameworkReconciler(client, bgReconciler, enabled)}
}
```

Syulin7: Done, thanks for your suggestion.

Syulin7:
@tenzen-y This PR is ready for review. PTAL.

pkg/controller.v1/common/job_controller.go
```diff
@@ -177,13 +178,27 @@ var GenSchedulerPluginsSetupFunc = func(c client.Client) GangSchedulingSetupFunc
...
 	}
 }
 
+var GenKoordinatorSetupFunc = func(c client.Client) GangSchedulingSetupFunc {
+	return func(jc *JobController) {
+		jc.Config.GangScheduling = GangSchedulerKoordinator
+		jc.PodGroupControl = control.NewKoordinatorControl(c)
+	}
+}
+
```
tenzen-y, line 201: Why are there the same functions (L181-L187)?

tenzen-y:
Can you squash commits into one?

Syulin7:
> Can you squash commits into one?

Done

pkg/controller.v1/common/job_controller.go
Syulin7: Done

tenzen-y (commented):
/lgtm
/assign @johnugeorge

johnugeorge:
@Syulin7 We will get back to this in a week. Sorry for the delay as we haven't released 1.6 release yet.

alculquicondor:
IMO, this is not sustainable. We'll keep adding support for more schedulers in the ecosystem?

schedulers need to look into generic approaches, such as the suspend field.

tenzen-y:
> IMO, this is not sustainable. We'll keep adding support for more schedulers in the ecosystem?
> 
> schedulers need to look into generic approaches, such as the suspend field.

@alculquicondor We don't have any criteria to add more schedulers. So we may need to have the criteria (e.g., only CNCF project?)

alculquicondor:
An alternative that I'm proposing in kubeflow/mpi-operator#500 is that koordinator implements this on a controller in their repo.

I don't see why that's not possible.

tenzen-y:
> An alternative that I'm proposing in [kubeflow/mpi-operator#500](https://github.com/kubeflow/mpi-operator/issues/500) is that koordinator implements this on a controller in their repo.
> 
> I don't see why that's not possible.

It looks like you have a point.
However, this issue applies to volcano and scheduler-plugins as well, not just to the koordinator.

alculquicondor:
If this works out well, it could be a precedent for justifying https://github.com/kubernetes/kubernetes/issues/107294, which all schedulers can use to generically group any object.

tenzen-y:
> If this works out well, we could justify:
> 
> 1. Adding similar fields to kubeflow's `schedulingPolicy` in `batch/v1.Job`
> 2. [Add suspend subresource kubernetes/kubernetes#107294](https://github.com/kubernetes/kubernetes/issues/107294), which all schedulers can use to generically group pods from a custom job.

Totally, I agree. However, since we may need long terms to complete all steps, I would like to add coscheduling support to MPIJob for the short term.

tenzen-y:
> Yes, it applies to all.
> 
> If the maintainers of kubeflow agree, we would have to:
> 
> 1. Open issues in volcano and scheduling-plugins to implement the logic in their side.
> 2. Make it possible to disable the PodGroup control in kubeflow (I believe this is possible by just leaving the `gang-scheduler` parameter empty).
> 3. Remove the code after a few releases.

Ah, I missed the above comments. I'm ok with having the logic for PodGroup in the scheduler-plugins repo.
If the scheduler-plugins team agrees with this, I can help to implement the logic on their repo.

alculquicondor:
For the time being, I would vote for only keeping support for the schedulers we already have (scheduler-plugins and volcano).

I would encourage the maintainers of Koordinator to put their efforts in https://github.com/kubernetes/kubernetes/issues/107294 and https://github.com/kubernetes/enhancements/issues/3370 instead of adding more dependencies in kubeflow.

Syulin7:
@alculquicondor  I understand your concerns, and it make sense. As I mentioned in https://github.com/kubeflow/training-operator/issues/1746, the Koordinator scheduler uses the PodGroup defined in scheduler-plugins/coscheduling. Therefore, supporting Koordinator in the training-operator will not add more dependencies, but rather provide another option. Currently, some people are using both the training-operator and Koordinator simultaneously, which is very useful for them.

In the long run, I think it's necessary to provide a Native PodGroup API that includes fields such as suspend, queue, minNumOfWorkers, etc. Different schedulers can implement the same Native PodGroup API. For example, the operator only needs to add a specific label to the pod, and the scheduler creates and maintains the PodGroup. This way, the operator and scheduler won't depend on each other. I think we can work towards this direction in the future, but supporting Koordinator in the training-operator currently will not affect this.

@alculquicondor @tenzen-y @terrytangyuan  WDYT

tenzen-y:
> Therefore, supporting Koordinator in the training-operator will not add more dependencies, but rather provide another option. 

That's right for now. However, the training-operator must directly depend on the koordinator if the koordinator-scheduler has original logic or CRDs in the future.

However, fortunately, the users can deploy CustomJobs with koordinator-scheduler by specifying scheduler-name even if we don't add this feature to the master branch as I mentioned in https://github.com/kubeflow/training-operator/issues/1746#issuecomment-1408364031.

So I agree with @alculquicondor.

pkg/controller.v1/common/job.go
```diff
@@ -275,7 +275,7 @@ func (jc *JobController) ReconcileJobs(
 					pg = volcanoPodGroup
 					return nil
 				}
-			case GangSchedulerSchedulerPlugins:
+			case GangSchedulerSchedulerPlugins, GangSchedulerKoordinator:
```
alculquicondor, line 278: why is the PR bigger than just this line?

if we can have this, I would support this change. Otherwise, I'm opposed.

```golang
switch jc.Config.GangScheduling {
case GangSchedulerVolcano:
  ...
default:
}
```

pkg/controller.v1/control/podgroup_control.go
```diff
@@ -193,3 +193,29 @@ func (s *SchedulerPluginsControl) CreatePodGroup(podGroup client.Object) error {
...
+
+func (s *KoordinatorControl) DecoratePodTemplateSpec(pts *corev1.PodTemplateSpec, job metav1.Object, _ string) {
+	if len(pts.Spec.SchedulerName) == 0 {
+		pts.Spec.SchedulerName = s.GetSchedulerName()
+	}
+	if pts.Labels == nil {
+		pts.Labels = make(map[string]string)
+	}
+	pts.Labels[schedulerpluginsv1alpha1.PodGroupLabel] = job.GetName()
+}
```
tenzen-y, line 210: > why is the PR bigger than just this line?
> 
> if we can have this, I would support this change. Otherwise, I'm opposed.

@alculquicondor We need to add this function if we natively support the koordinator-scheduler.
So, you may oppose this change.

pkg/controller.v1/common/job.go
alculquicondor: here, I'm considering scheduling-plugins the "default" because it has a higher chance of becoming part of upstream kubernetes at some point.

pkg/controller.v1/control/podgroup_control.go
alculquicondor: This is no different from SchedulerPluginsControl. They can remove `KoordinatorControl` and use the existing implementation.

tenzen-y: Sorry for your confusion.
We must handle another value in the function GetSchedulerName (L212 - L214 in this file).

This means we must specify the `koord-scheduler` to the schedulerName in podSpec.

alculquicondor: if you treat `SchedulerPluginsControl` as the default implementation when the gang-scheduler-plugin configuration is set, you can pass that.

Syulin7: > if you treat `SchedulerPluginsControl` as the default implementation when the gang-scheduler-plugin configuration is set, you can pass that string, even if it's not "scheduler-plugins"

I think this is ok,add --enable-gang-scheduler and --gang-scheduler-name. SchedulerPluginsControl as the default implementation when the enable-gang-scheduler is true, and set gang-scheduler-name if it is not nil.

alculquicondor: just not empty gang-scheduler-name should be enough. No need for an extra enable-gang-scheduler

alculquicondor: sgtm

tenzen-y: > scheduler-plugins as secondary scheduler, set gang-scheduler-name=scheduler-plugins(or any other name they use. like koordinator.) scheduler-plugins as default scheduler, set gang-scheduler-name=default-scheduler volcano as secondary scheduler, set gang-scheduler-name=volcano @tenzen-y @alculquicondor WDYT

@Syulin7 I have a question. If we set gang-scheduler-name=default-scheduler, does the training-operator embed `default-scheduler` in podTemplateSpec.spec.schedulerName? Or does the training-operator skip set a value in podTemplateSpec.spec.schedulerName?

Syulin7: @tenzen-y  I think if we set gang-scheduler-name=default-scheduler, the training-operator should skip set a value in podTemplateSpec.spec.schedulerName.

tenzen-y: sgtm

Syulin7:
@tenzen-y @alculquicondor  This PR is ready for review. PTAL.

I set SchedulerPluginsControl as the default implementation. 
Removed koordinator related code, set --gang-scheduler-name=koord-scheduler can support Koordinator Gang Scheduler.

The test results are as follows.
```
$ kubectl get pods -n kubeflow
NAME                    READY   STATUS    RESTARTS   AGE
tfjob-simple-worker-0   0/1     Pending   0          82s
tfjob-simple-worker-1   0/1     Pending   0          82s
$ kubectl describe pods -n kubeflow tfjob-simple-worker-0 | grep -A10  Events
Events:
  Type     Reason            Age   From             Message
  ----     ------            ----  ----             -------
  Warning  FailedScheduling  16m   koord-scheduler  running PreFilter plugin "Coscheduling": gang child pod not collect enough, gangName: kubeflow/tfjob-simple, podName: kubeflow/tfjob-simple-worker-0
  Warning  FailedScheduling  16m   koord-scheduler  running PreFilter plugin "Coscheduling": gang child pod not collect enough, gangName: kubeflow/tfjob-simple, podName: kubeflow/tfjob-simple-worker-0
```

pkg/controller.v1/common/job_controller.go
```diff
@@ -170,16 +170,15 @@ var GenVolcanoSetupFunc = func(vci volcanoclient.Interface) GangSchedulingSetupF
...
-		jc.Config.GangScheduling = GangSchedulerSchedulerPlugins
-		jc.PodGroupControl = control.NewSchedulerPluginsControl(c)
+		jc.Config.GangScheduling = GangScheduler(gangSchedulerName)
+		jc.PodGroupControl = control.NewSchedulerPluginsControl(c, gangSchedulerName)
 	}
 }
 
 var GenNonGangSchedulerSetupFunc = func() GangSchedulingSetupFunc {
 	return func(jc *JobController) {
-		jc.Config.GangScheduling = GangSchedulerNone
```
tenzen-y, line 182: Should we enforce an empty string? 

```go
jc.Config.GangScheduling = ""
```

pkg/controller.v1/control/podgroup_control.go
```diff
@@ -191,5 +196,3 @@ func (s *SchedulerPluginsControl) CreatePodGroup(podGroup client.Object) error {
 	}
 	return nil
 }
-
-var _ PodGroupControlInterface = &SchedulerPluginsControl{}
```
tenzen-y, line 195: Why is this removed?

pkg/controller.v1/common/job_controller.go
```diff
@@ -69,9 +69,9 @@ var (
 type GangScheduler string
 
 const (
-	GangSchedulerNone             GangScheduler = "None"
 	GangSchedulerVolcano          GangScheduler = "volcano"
 	GangSchedulerSchedulerPlugins GangScheduler = "scheduler-plugins"
+	GangSchedulerDefault          GangScheduler = "default-scheduler"
```
alculquicondor, line 74: Add a comment that using this scheduler name or any scheduler name different than volcano uses the scheduler-plugins PodGroup

pkg/controller.v1/common/job_controller.go
```diff
@@ -69,9 +69,9 @@ var (
 type GangScheduler string
 
 const (
-	GangSchedulerNone             GangScheduler = "None"
```
alculquicondor, line 72: oh I wasn't aware that this existed. Maybe we can add another:

```
case GangSchedulerNone, "":
  return nil
default:
  // scheduler plugins
```

pkg/controller.v1/common/pod.go
```diff
@@ -421,7 +421,7 @@ func (jc *JobController) createNewPod(job interface{}, rt string, index int, spe
 
 	// if gang-scheduling is enabled:
 	// 1. if user has specified other scheduler, we report a warning without overriding any fields.
-	// 2. if no SchedulerName is set for pods, then we set the SchedulerName to "volcano".
+	// 2. if no SchedulerName is set for pods, then we set the SchedulerName to enabled scheduler name.
```
alculquicondor, line 424: was this assuming that the default scheduler was volcano? Or is the comment innacurate?

pkg/controller.v1/control/podgroup_control.go
```diff
@@ -125,23 +125,28 @@ var _ PodGroupControlInterface = &VolcanoControl{}
 
 // SchedulerPluginsControl is the  implementation of PodGroupControlInterface with scheduler-plugins.
 type SchedulerPluginsControl struct {
-	Client client.Client
+	Client        client.Client
+	SchedulerName string
 }
 
 func (s *SchedulerPluginsControl) DecoratePodTemplateSpec(pts *corev1.PodTemplateSpec, job metav1.Object, _ string) {
+	if len(pts.Spec.SchedulerName) == 0 && s.GetSchedulerName() != "default-scheduler" {
```
alculquicondor, line 133: why check `GetSchedulerName != "default-schedule"`?

pkg/reconciler.v1/common/gang_scheduler_framework.go
```diff
@@ -38,6 +38,7 @@ type SchedulerFrameworkReconciler struct {
 	BaseGangReconciler
 	ReconcilerUtilInterface
 	client.Client
+	SchedulerName string
 }
 
 func BareSchedulerFrameworkReconciler(client client.Client, bgReconciler *BaseGangReconciler, enabled bool) *SchedulerFrameworkReconciler {
```
alculquicondor, line 44: What is this for @tenzen-y ?

tenzen-y: Maybe, this function is expected using in tests like the following:

https://github.com/kubeflow/common/blob/4dbf18c85417b54158914d1d01c1744b4e3542f2/test_job/reconciler.v1/test_job/test_job_reconciler.go#L55

tenzen-y: However, we don't have any test reconciler for scheduler-plugins for now.

alculquicondor: so it basically schedules the pods?

tenzen-y: Ah, I remind the purpose of this function.
This function isn't only for tests.

This just creates a new instance for the `SchedulerFrameworkReconciler`.

However, we use controllers instead of reconcilers. So we don't use this package for now.
More details:  #140

pkg/controller.v1/common/pod.go
tenzen-y: > was this assuming that the default scheduler was volcano?

We used to decorate the schedulerName in podTemplateSpec.spec.schedulerName only when the gang-scheduler name is set as `volcano`.

- volcano

https://github.com/kubeflow/common/blob/4dbf18c85417b54158914d1d01c1744b4e3542f2/pkg/controller.v1/control/podgroup_control.go#L65-L74

- scheduler-plugins

https://github.com/kubeflow/common/blob/4dbf18c85417b54158914d1d01c1744b4e3542f2/pkg/controller.v1/control/podgroup_control.go#L131-L136

alculquicondor: so the comment was inaccurate, right?

pkg/reconciler.v1/common/gang_scheduler_framework.go
Syulin7: > However, we use controllers instead of reconcilers. So we don't use this package for now. More details: #140
yes, we don't use this package for now.

pkg/controller.v1/control/podgroup_control.go
Syulin7: Done. My mistake

pkg/controller.v1/common/job_controller.go
Syulin7: Done

pkg/controller.v1/common/job_controller.go
```diff
@@ -170,16 +170,15 @@ var GenVolcanoSetupFunc = func(vci volcanoclient.Interface) GangSchedulingSetupF
 	}
 }
 
-var GenSchedulerPluginsSetupFunc = func(c client.Client) GangSchedulingSetupFunc {
+var GenSchedulerPluginsSetupFunc = func(c client.Client, gangSchedulerName string) GangSchedulingSetupFunc {
```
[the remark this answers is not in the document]
Syulin7: it is used in training-operator: https://github.com/kubeflow/training-operator/pull/1747/files#diff-900bf2495e0ccd04b3bdbd8cfabea9ee5eb73a6e3982a0a928bd6ed6e4ebe2d6L124-L125

pkg/controller.v1/control/podgroup_control.go
Syulin7: if gang-scheduler-name=default-scheduler, skip set a schedulerName in podTemplateSpec.spec.schedulerName.

if gang-scheduler-name is set to a different value(e.g. scheduler-plugins, koordinator), we should set the schedulerName in podTemplateSpec.spec.schedulerName.

pkg/controller.v1/common/job_controller.go
Syulin7: Done

pkg/controller.v1/common/job_controller.go
Syulin7: Done

pkg/controller.v1/common/pod.go
Syulin7: I update the comment.

Syulin7:
I resolved all conversations, PTAL, thanks. @tenzen-y @alculquicondor

pkg/controller.v1/common/job.go
```diff
@@ -275,7 +275,9 @@ func (jc *JobController) ReconcileJobs(
 					pg = volcanoPodGroup
 					return nil
 				}
-			case GangSchedulerSchedulerPlugins:
+			case GangSchedulerNone:
+				return nil
```
tenzen-y, line 279: Probably, we don't need this condition since we confirm if the JobConfig has a config for gang scheduling in L232.

cc: @alculquicondor

pkg/controller.v1/common/job_controller.go
```diff
@@ -69,9 +69,11 @@ var (
 type GangScheduler string
 
 const (
-	GangSchedulerNone             GangScheduler = "None"
+	GangSchedulerNone             GangScheduler = ""
```
tenzen-y, line 72: What intent of using an empty string instead of `None`?

Syulin7: if GangSchedulerNone is None, It may be considered as enabling the gang scheduling, but I think we can remove the GangSchedulerNone const.
```
var GenNonGangSchedulerSetupFunc = func() GangSchedulingSetupFunc {
	return func(jc *JobController) {
		jc.Config.GangScheduling = GangSchedulerNone
		jc.PodGroupControl = nil
	}
}

func (c *JobControllerConfiguration) EnableGangScheduling() bool {
	return c.GangScheduling != ""
}
```

pkg/controller.v1/common/job.go
Syulin7: Done

pkg/controller.v1/common/job_controller.go
tenzen-y: Thanks for clarifying.

>  but I think we can remove the GangSchedulerNone const.

Sounds good. However, it might be better to do that after we merge kubeflow/common into kubeflow/training-operator since GangSchedulerNone might be worth it as documentation.

Syulin7: > since GangSchedulerNone might be worth it as documentation.

Make sense, let's keep it.

pkg/controller.v1/common/job_controller.go
```diff
@@ -81,7 +83,7 @@ type JobControllerConfiguration struct {
 }
 
 func (c *JobControllerConfiguration) EnableGangScheduling() bool {
-	return c.GangScheduling != GangSchedulerNone
+	return c.GangScheduling != ""
```
alculquicondor, line 86:
```suggestion
	return c.GangScheduling != "" and c.GangScheduling != GangSchedulerNone
```

pkg/controller.v1/control/podgroup_control.go
alculquicondor: why not just always set?

pkg/controller.v1/common/job_controller.go
tenzen-y: Yes, the "None" was already released on v0.4.6.

Syulin7: Yes, we should keep it for backwards compatibility.

pkg/controller.v1/common/job_controller.go
```diff
@@ -72,6 +72,8 @@ const (
 	GangSchedulerNone             GangScheduler = "None"
 	GangSchedulerVolcano          GangScheduler = "volcano"
 	GangSchedulerSchedulerPlugins GangScheduler = "scheduler-plugins"
+	// GangSchedulerDefault Using this scheduler name or any scheduler name different than volcano uses the scheduler-plugins PodGroup
+	GangSchedulerDefault GangScheduler = "default-scheduler"
```
alculquicondor, line 76: Remove if this is not being used

alculquicondor (commented):
a nit, otherwise lgtm

alculquicondor:
/retitle Make scheduler-plugins the default gang scheduler

pkg/reconciler.v1/common/gang_scheduler_framework.go
```diff
@@ -164,6 +165,10 @@ func (r *SchedulerFrameworkReconciler) DecoratePodForGangScheduling(
 	podTemplate *corev1.PodTemplateSpec,
 	job client.Object,
 ) {
+	if podTemplate.Spec.SchedulerName == "" && r.GetGangSchedulerName() != "default-scheduler" {
```
tenzen-y, line 168:
```suggestion
	if podTemplate.Spec.SchedulerName == "" {
```

Let's use a logic the same as the podgroup_control.

tenzen-y:
I left a comment for nit. Overall, lgtm.

Syulin7:
> I left a comment for nit. Overall, lgtm.

tenzen-y:
@Syulin7 Thanks for the great work!
/lgtm

@johnugeorge 
Do you decide on the week for releasing the training operator v1.6, although releasing kubeflow v1.7 seems delayed?

> Week 23 - Distribution Testing Ends: Wednesday, Mar 22nd 2023

pkg/controller.v1/control/podgroup_control.go
```diff
@@ -125,23 +125,28 @@ var _ PodGroupControlInterface = &VolcanoControl{}
...
 
 func (s *SchedulerPluginsControl) GetSchedulerName() string {
-	return ""
+	return s.SchedulerName
 }
 
 // NewSchedulerPluginsControl returns a SchedulerPluginsControl
-func NewSchedulerPluginsControl(c client.Client) PodGroupControlInterface {
-	return &SchedulerPluginsControl{Client: c}
+func NewSchedulerPluginsControl(c client.Client, schedulerName string) PodGroupControlInterface {
```
alculquicondor, line 148: Not about this PR, but I realized that some functions below are unnecessary, like `UpdatePodGroup`, and `CreatePodGroup`. They are casting unnecessarily: the controller-runtime client can already handle the casting via reflection. And I think `GetPodGroup` and `DeletePodGroup` could be refactored so that we don't need the casting either. Maybe even use generics.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

johnugeorge:
@tenzen-y The final release is not yet made due to delay in Kubeflow upstream. Sorry for the delay. This should not happen again in the future as we always branch out for all repos except common.  We will target this week end to merge- Friday 24 March

tenzen-y:
> @tenzen-y The final release is not yet made due to delay in Kubeflow upstream. Sorry for the delay. This should not happen again in the future as we always branch out for all repos except common. We will target this week end to merge- Friday 24 March

Thanks for letting me know! I appreciate your hard work!

johnugeorge:
Training operator release is made. This can be merged.

Thanks for the patience

tenzen-y:
@Syulin7 Thank you!
/hold cancel

tenzen-y:
@johnugeorge Can we cut a new release on this repo to move forward https://github.com/kubeflow/training-operator/pull/1747?

PR #214, opened 2023-04-21
fix job active count;  count terminating pod as failed

yowenter:
if pod is terminating , the pod.Status.Phase is still RUNNING. 
In this case, the Job status maybe not correct.

### DIFF
--- a/pkg/controller.v1/common/status_test.go
+++ b/pkg/controller.v1/common/status_test.go
@@ -2,10 +2,12 @@ package common
 
 import (
 	"testing"
+	"time"
 
 	apiv1 "github.com/kubeflow/common/pkg/apis/common/v1"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestUpdateJobReplicaStatuses(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/core/status.go
```diff
@@ -34,7 +34,14 @@ func InitializeReplicaStatuses(jobStatus *apiv1.JobStatus, rtype apiv1.ReplicaTy
...
 	case corev1.PodRunning:
-		jobStatus.ReplicaStatuses[rtype].Active++
+		if pod.DeletionTimestamp != nil {
+			// when node is not ready, the pod will be in terminating state.
+			// Count deleted Pods as failures to account for orphan Pods that
+			// never have a chance to reach the Failed phase.
+			jobStatus.ReplicaStatuses[rtype].Failed++
+		} else {
+			jobStatus.ReplicaStatuses[rtype].Active++
+		}
```
tenzen-y, line 44: I don't think that the deleting pod is failed one. Please let me know if I'm missing any cases.

yowenter: Hi, @tenzen-y, 

I found some code in `pkg/util/k8sutil/k8sutil.go` 
```
func IsPodActive(p *v1.Pod) bool {
	return v1.PodSucceeded != p.Status.Phase &&
		v1.PodFailed != p.Status.Phase &&
		p.DeletionTimestamp == nil
}
```

, it seems that job active count should not contains the pod in `Terminating` state.
And in kubernetes jobs implementation, the `Terminating` pod doesn't count as active either.
In some cases, if the node becomes not ready, we are not certain if  the pod is RUNNING or not...

tenzen-y: Uhm. You have a point. When the pod is running and has a DeletionTimestamp field, we should mark Failed to the Job.

Thank you for clarifying!

tenzen-y:
@yowenter And also, can you update the unit test?

https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/status_test.go

yowenter:
> @yowenter And also, can you update the unit test?
> 
> https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/status_test.go

Hi, @tenzen-y,  I've added job status unittest, please review again.

pkg/controller.v1/common/status_test.go
```diff
@@ -2,10 +2,13 @@ package common
 
 import (
 	"testing"
+	"time"
 
 	apiv1 "github.com/kubeflow/common/pkg/apis/common/v1"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+
```
tenzen-y: nit: revert this line.

pkg/controller.v1/common/status_test.go
```diff
@@ -14,13 +17,14 @@ func TestUpdateJobReplicaStatuses(t *testing.T) {
 	_, ok := jobStatus.ReplicaStatuses["worker"]
 	// assert ReplicaStatus for "worker" exists
 	assert.True(t, ok)
-	setStatusForTest(&jobStatus, "worker", 2, 3, 1)
-	assert.Equal(t, jobStatus.ReplicaStatuses["worker"].Failed, int32(2))
+	setStatusForTest(&jobStatus, "worker", 2, 3, 1, 1)
+	// terminating pod should count as failed.
```
tenzen-y, line 20: good comment :)

tenzen-y (commented):
@yowenter Looks great! Thank you!

/lgtm
/assign @terrytangyuan

tenzen-y:
/lgtm cancel 
for nit

terrytangyuan (approved):
/approve

